PR #129, opened 2026-03-18
feat(activerecord): unskip 9 base.test.ts tests

deanmarano:
## Summary

This picks up the first chunk of workstream B (PR B1) from the activerecord 100% plan. I went through the 64 skipped tests in `base.test.ts` and implemented bodies for the 9 that don't require major new infrastructure (time zones, multi-db connections, marshalling, eager loading, etc.).

The unskipped tests cover:
- `attributesBeforeTypeCast` / `readAttributeBeforeTypeCast` (which already existed on Model but the base.test.ts test was still stubbed)
- `respondTo` for nonexistent methods
- Column names with spaces
- Array quoting in WHERE clauses
- Unicode string handling
- Out-of-range and array-based `find` with slugs
- Datetime attribute assignment (valid and invalid values)

The remaining 55 skipped tests are blocked on features that belong to other PRs — time zone infrastructure, multi-database connections (B8), eager loading (A2), associations (A*), marshal/benchmark (Ruby-only), etc.

Results: 237 -> 246 passing, 64 -> 55 skipped.

### DIFF
--- a/packages/activerecord/src/base.test.ts
+++ b/packages/activerecord/src/base.test.ts
@@ -1755,6 +1755,8 @@ describe("BasicsTest", () => {
     expect(results).toHaveLength(2);
     expect(results[0].readAttribute("title")).toBe("first");
     expect(results[1].readAttribute("title")).toBe("second");
+    const reversed = await Topic.find([`${t2.id}-hello`, `${t1.id}-meowmeow`]);
+    expect(reversed[0].readAttribute("title")).toBe("second");
   });
   it.skip("find by slug with range", () => {});
   it.skip("equality of relation and collection proxy", () => {});
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -833,15 +833,12 @@ export class Base extends Model {
    *
    * Mirrors: ActiveRecord::Base.find
    */
-  /**
-   * Cast a value through the primary key's type, mimicking Rails' id coercion.
-   * For integer PKs, "1-meowmeow".to_i → 1 in Ruby; parseInt("1-meowmeow") → 1 in JS.
-   */
-  private static _castPrimaryKeyValue(value: unknown): unknown {
-    if (typeof this.primaryKey !== "string") return value;
-    const def = this._attributeDefinitions.get(this.primaryKey);
+  /** @internal Cast a value through an attribute's type, with parseInt fallback for the default PK. */
+  static _castAttributeValue(key: string, value: unknown): unknown {
+    if (typeof value !== "string") return value;
+    const def = this._attributeDefinitions.get(key);
     if (def) return def.type.cast(value);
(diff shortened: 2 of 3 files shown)


deanmarano (commented):
## Re-review (after updates)

Nice work addressing the feedback. This is much improved.

### Fixed tests

**out of range slugs** -- Now properly tests slug parsing with `"${t1.id}-meowmeow"` and the out-of-range `"9223372036854775808-hello"`. Matches Rails intent well.

**find by slug with array** -- Now uses `Topic.find(["${t1.id}-meowmeow", "${t2.id}-hello"])`. Properly tests slug parsing in array context. One minor gap vs Rails: the Rails test also checks `Topic.find(["2-hello", "1-meowmeow"]).first.title` to verify ordering is preserved when slugs are in non-id order.

**respect internal encoding** -- Good call re-skipping with a clear explanation.

**attributes on dummy time / invalid time** -- Good call re-skipping both with notes about what's blocking.

**non valid identifier column name** -- Now uses `a$b` (correct character) and includes create + find roundtrip. Missing the `update_columns` + reload second half from Rails, but the core intent is covered.

**quoting arrays** -- Now executes the query end-to-end and includes the empty-array edge case. Matches Rails intent.

**attributes_before_type_cast** -- Added `attributesBeforeTypeCast` assertion too. Looks solid.

### Implementation changes (base.ts, relation.ts)

The `_castPrimaryKeyValue` method for slug parsing is clean. Using `parseInt` to mimic Ruby's `.to_i` is a good analog -- both extract the leading integer from a string.

The `_castWhereValue` in Relation is a reasonable approach for casting where clause values through the attribute type system. A few observations:

1. **`_castWhereValue` only handles strings** -- it returns early for non-string values. This is fine for slug parsing but worth noting that Ruby's type casting is broader (e.g., casting `"42"` to integer in any where clause). Seems intentional given the scope of this PR though.

2. **The PK fallback in `_castWhereValue`** does `parseInt` when there's no attribute definition but the key matches the primary key. This duplicates the logic in `_castPrimaryKeyValue`. Minor, but could drift.

3. **`_castWhereValue` is called for every where value** including in `whereNot`. Good consistency.

### Still missing from Rails tests

- `attributes_before_type_cast` and `raise no method error for nonexistent method` still don't correspond to tests in Rails `base_test.rb`. Are these from a different test file, or were they added independently? If they're not matched by `convention:compare`, that's fine, but worth confirming.

- `find by slug with array` doesn't test that result ordering matches input order (Rails line 590: `Topic.find(["2-hello", "1-meowmeow"]).first.title`).

### Overall

This is in good shape now. The remaining unskipped tests (5) faithfully test what their names describe, the re-skipped tests have clear explanations, and the implementation changes for slug casting are clean. The only thing I'd want to see addressed is confirming the two tests that don't exist in Rails base_test.rb.

PR #147, opened 2026-03-20
feat(activerecord): through association joins and write protection

deanmarano:
## Summary

This picks up the A1 workstream by implementing through association join generation and write protection enforcement -- the two most impactful pieces from the remaining through-association work.

The main changes:

- **Through association joins**: 9 new tests unskipped across `has-many-through-associations.test.ts` covering `joins`, `left_joins`, `distinct`, belongs_to-through, explicit join table joining, merge with proxy, sum on columns, and inner join with quoted table names. These all exercise the existing `_resolveAssociationJoin` / `_resolveThroughJoin` infrastructure to verify it generates correct SQL for various through patterns.

- **Write protection**: CollectionProxy now guards write operations (push, delete, replace) against two read-only scenarios that Rails enforces:
  - Through a `has_one` raises `HasManyThroughCantAssociateThroughHasOneOrManyReflection`
  - Nested through (through another through) raises `HasManyThroughNestedAssociationsAreReadonly` / `HasOneThroughNestedAssociationsAreReadonly`

- **Table-qualified where keys**: `where({ "posts.title": "active" })` now correctly generates `"posts"."title" = 'active'` instead of treating the whole dotted string as a column name. This is pretty important for join queries where you need to disambiguate columns across tables. Fixed in all three SQL generation paths (`_buildWhereNodes`, `_buildWhereStrings`, `_applyWheresToManager`).

Convention:compare goes from 52.2% to 56.7% (4754/8385 tests).

### DIFF
--- a/packages/activerecord/src/relation.ts
+++ b/packages/activerecord/src/relation.ts
@@ -2014,7 +2014,7 @@ export class Relation<T extends Base> {
       }
     }
     for (const rawJoin of this._rawJoins) {
-      manager.join(new Nodes.StringJoin(new Nodes.SqlLiteral(rawJoin)));
+      (manager as any).core.source.right.push(new Nodes.StringJoin(new Nodes.SqlLiteral(rawJoin)));
     }
   }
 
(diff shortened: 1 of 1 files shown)


deanmarano:
Re: the `HasManyThroughCantAssociateThroughHasOneOrManyReflection` error message mismatch — this error class name and message are taken verbatim from Rails. The "has_one or has_many" in the message refers to the source reflection type on the through model. Our check guards against `hasOne` through associations (where writes should be blocked), while `hasMany` through is the normal write-allowed pattern.

PR #488, opened 2026-04-10
feat(stats): backfill workflow steps for historical jobs

deanmarano:
## Summary
- Adds `syncWorkflowSteps` backfill pass that finds runs whose jobs are missing steps and re-fetches the jobs API (which returns steps inline)
- Wired into main() after `syncCheckAnnotations`

## Context
When workflow runs were synced before #485, jobs were stored without their `steps` array. This pass populates them for the ~2500 historical jobs that were missed.

## Test plan
- [ ] Run \`pnpm stats:sync --refresh\` and verify workflow_steps count grows

### DIFF
--- a/scripts/sync-stats/sync.ts
+++ b/scripts/sync-stats/sync.ts
@@ -1580,6 +1580,64 @@ async function syncCheckAnnotations(mode: "latest" | "refresh") {
   }
 }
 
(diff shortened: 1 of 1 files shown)


deanmarano:
Keeping changes on main for local testing instead.

PR #558, opened 2026-04-16
feat(activerecord): implement all missing PostgreSQL quoting methods

deanmarano:
## Summary

Rebased onto main. PR #560 delivered most of the PG quoting surface (abstract delegation, OID wrapper types, BinaryBind typeCast shape, quoteDefaultExpression with type map), so this PR is reduced to the six remaining methods needed for full Rails parity:

- **quoteTableNameForAssignment** — returns the quoted column name (PG drops the table prefix on `UPDATE ... SET`).
- **quoteSchemaName** — delegates to `quoteColumnName`.
- **quotedBinary** — `'#{escape_bytea(value.to_s)}'`.
- **checkIntInRange** — Rails' method name, aliased to the existing `checkIntegerRange`.
- **columnNameWithOrderMatcher** — column matcher + `COLLATE "…"`/`ASC`/`DESC`/`NULLS FIRST|LAST`. Rails-faithful: only quoted collation names are accepted.
- **lookupCastTypeFromColumn** — forwards `oid`/`fmod`/`sqlType` to the type map.

Each method mirrors Rails exactly — references the Rails source in the doc comment.

## Test plan

- [x] 17 quoting tests pass, including new cases for each added method
- [x] Full typecheck clean

### DIFF
--- a/packages/activerecord/src/connection-adapters/postgresql/quoting.test.ts
+++ b/packages/activerecord/src/connection-adapters/postgresql/quoting.test.ts
@@ -5,11 +5,18 @@ import { Data as BitData } from "./oid/bit.js";
 import { Range } from "./oid/range.js";
 import { Data as XmlData } from "./oid/xml.js";
 import {
+  checkIntInRange,
   columnNameMatcher,
+  columnNameWithOrderMatcher,
+  IntegerOutOf64BitRange,
+  lookupCastTypeFromColumn,
   quote,
   quoteDefaultExpression,
+  quotedBinary,
   quotedFalse,
   quotedTrue,
+  quoteSchemaName,
+  quoteTableNameForAssignment,
   typeCast,
   unescapeBytea,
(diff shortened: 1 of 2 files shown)


deanmarano:
Reviewed the 3 new Copilot suggestions — all three would diverge from Rails, so skipping:

1. **Unquoted COLLATE** — Rails' `column_name_with_order_matcher` uses `(?:\s+COLLATE\s+"\w+")?` (quoted only). Current impl matches.
2. **uuid `value.includes("()")`** — Rails does `value.include?("()")` verbatim. Current impl matches.
3. **`typeCast` binary shape** — Rails returns `{ value: value.to_s, format: 1 }` for `Type::Binary::Data`; we mirror that. `typeCast` isn't wired into any bind path yet — when it is, the PG adapter will need to unwrap the hash, but changing the return shape would break Rails parity.

PR #560, opened 2026-04-16
feat(activerecord): flesh out PostgreSQL OID core

deanmarano:
## Summary
- expand PostgreSQL OID::Array with Rails-like Data serialization, schema formatting, mapping, change detection, and force-equality behavior
- expand PostgreSQL OID::Range while preserving existing query Range usage
- align TypeMapInitializer with Rails run/query condition APIs and subtype registration for arrays, ranges, enums, domains, mapped types, and composites
- align PostgreSQL quoting edge cases for BinaryData bind shape, PG column default serialization, Xml/Bit/Array/Range/non-finite numeric quoting, and bytea helpers
- add focused tests for Array, Range, TypeMapInitializer, Vector wiring, and PostgreSQL quoting

## Validation
- pnpm vitest run packages/activerecord/src/connection-adapters/postgresql/quoting.test.ts
- pnpm vitest run packages/activerecord/src/connection-adapters/postgresql/oid/array.test.ts packages/activerecord/src/connection-adapters/postgresql/oid/range.test.ts packages/activerecord/src/connection-adapters/postgresql/oid/type-map-initializer.test.ts
- pnpm vitest run packages/activerecord/src/connection-adapters/postgresql/oid/array.test.ts packages/activerecord/src/connection-adapters/postgresql/oid/range.test.ts packages/activerecord/src/connection-adapters/postgresql/oid/type-map-initializer.test.ts packages/activerecord/src/adapters/postgresql/array.test.ts packages/activerecord/src/adapters/postgresql/range.test.ts packages/activerecord/src/adapters/postgresql/type-lookup.test.ts
- pnpm --filter @blazetrails/activerecord build
- pnpm api:compare --package activerecord (activerecord 2412/2859, 84.4%; OID array/range/type_map_initializer/vector remain 100%; PostgreSQL quoting now 11/17)
- pnpm test:compare --package activerecord

## Notes
- PostgreSQL quotedTrue/quotedFalse now emit inherited abstract SQL literals TRUE/FALSE instead of quoted t/f, matching Rails but changing emitted boolean literal SQL on existing code paths.
- Full pre-commit repo build is currently blocked by an existing unrelated TS4019 error in packages/activesupport/src/core-ext/object/try.ts: exported class Delegator implements private name Tryable.

### DIFF
--- a/packages/activerecord/src/connection-adapters/postgresql/oid/array.test.ts
+++ b/packages/activerecord/src/connection-adapters/postgresql/oid/array.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import { Array as OidArray, Data } from "./array.js";
 
 const stringSubtype = {
+  type: "string",
   cast: (value: unknown) => (value == null ? null : String(value)),
   serialize: (value: unknown) => (value == null ? null : String(value)),
   deserialize: (value: unknown) => (value == null ? null : String(value)),
@@ -9,6 +10,18 @@ const stringSubtype = {
 };
 
 describe("PostgreSQL::OID::Array", () => {
+  it("delegates type to the subtype", () => {
+    const type = new OidArray(stringSubtype);
+
+    expect(type.type).toBe("string");
+  });
+
+  it("casts scalar values through the subtype", () => {
+    const type = new OidArray(stringSubtype);
+
+    expect(type.cast(1)).toBe("1");
+  });
+
   it("serialize returns Data with encoder and casted values", () => {
     const type = new OidArray(stringSubtype);
     const data = type.serialize(["a", "b"]) as Data;
@@ -17,6 +30,12 @@ describe("PostgreSQL::OID::Array", () => {
     expect(data.encoder).toBe(type);
     expect(data.values).toEqual(["a", "b"]);
     expect(String(data)).toBe("{a,b}");
+  });
+
+  it("serialize returns non-array values unchanged", () => {
+    const type = new OidArray(stringSubtype);
+
+    expect(type.serialize("not an array")).toBe("not an array");
   });
 
   it("typeCastForSchema formats array elements through the subtype", () => {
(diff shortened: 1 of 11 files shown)


deanmarano:
Addressed the Array.deserialize routing bug (now calls subtype.deserialize when available, matching Rails' type_cast_array(..., :deserialize)).

Skipped the encodeRange bound-quoting suggestion — Rails' encode_range is literally:
`"[#{type_cast_range_value(range.begin)},#{type_cast_range_value(range.end)}#{...}"`

No per-bound quoting; the whole literal gets quoted by the outer quote()/quote_string. Adding double-quote escaping around each bound would diverge from Rails and produce different SQL output.

deanmarano:
Addressed both Copilot comments:

1. **Tryable interface+const duplicate** — false positive. TS declaration merging between an interface and a const of the same name is valid (they occupy separate namespaces, type vs value). Typecheck passes clean. Reverting to TryableMixin + re-export was what introduced the TS4019 'private name implements' error in the first place.

2. **assertValidRegistration NaN** — real bug, but Copilot's suggestion diverges from Rails. Rails does `oid.to_i`, which in Ruby returns **0** for non-numeric strings, not an error. The TS helper used `parseInt` which returns NaN. Fixed by updating `toInt` to mirror Ruby's `String#to_i` coercion — returns 0 for non-numeric input. Invalid OIDs now register under 0 matching Rails exactly (including Rails' subtly broken behavior where all invalid OIDs collide on 0, which is a Rails-upstream concern).

deanmarano:
Addressed 1 of 3 Copilot comments:

**#1 queryConditionsForKnownTypeNames filter** — skipped. Rails does `@store.keys.map { |n| "'#{n}'" }` with no filter and no empty guard. Our impl matches Rails verbatim (plus defensive quote-escape, already documented as a divergence). Filtering to string keys would diverge from Rails.

**#2 registerDomainType alias vs register** — skipped. Rails literally does `register row["oid"], base_type` after `@store.lookup(row["typbasetype"].to_i)`. Our code is a direct mirror. Switching to aliasType would diverge from Rails.

**#3 Array.deserialize nested arrays** — fixed in `5edd2d15`. Rails' deserialize only sees strings because PG::TextDecoder runs upstream. node-pg can decode arrays client-side before our code runs, so routing `Array.isArray` through `typeCastArray` (mirroring what `cast()` already does) handles the TS-specific reality. Annotated inline as a Rails divergence.

PR #568, opened 2026-04-17
refactor(activerecord): retrofit 7 PG OID classes to extend Type::Value

deanmarano:
## Summary

First batch toward Rails-faithful PG adapter type_map wiring. Rails' entire \`OID::\` hierarchy descends from \`ActiveModel::Type::Value\`; on main these classes were standalone, which meant they couldn't be registered in a \`HashLookupTypeMap\` or participate in the adapter's type resolution path.

Retrofits (inheritance matches Rails exactly):

- **OID::Bit** extends \`Type<string>\` — \`cast_value\` now handles \`0x...\` hex notation → binary string like Rails; \`serialize\` wraps in \`Bit::Data\` with \`isBinary()\` / \`isHex()\` predicates mirroring Rails \`Data#binary?\` / \`#hex?\`.
- **OID::BitVarying** extends \`Bit\` (Rails: \`class BitVarying < OID::Bit\`).
- **OID::Bytea** extends \`BinaryType\` (Rails: \`class Bytea < Type::Binary\`) — only overrides \`deserialize\` to handle \`\\x...\` hex escape, falls back to parent for raw bytes / Data.
- **OID::Cidr** extends \`Type<string>\` (Rails: \`class Cidr < Type::Value\`).
- **OID::Inet** extends \`Cidr\` (Rails: \`class Inet < Cidr\`).
- **OID::Macaddr** extends \`StringType\` (Rails: \`class Macaddr < Type::String\`) — \`isChanged\` / \`isChangedInPlace\` do case-insensitive comparison like Rails' \`casecmp\`, so casing differences don't mark columns dirty.
- **OID::Decimal** extends \`DecimalType\` (Rails: \`class Decimal < Type::Decimal\`) — adds \`infinity(options)\` returning ±Infinity (Rails returns \`BigDecimal(\"Infinity\")\`; JS has no BigDecimal analog).

\`HashLookupTypeMap\` now accepts numeric keys so it can hold the OID-int registrations that layer on top of typname-string registrations, matching Rails' Hash semantics.

**Follow-up:** next PR retrofits \`Jsonb\`, \`Xml\`, \`Money\`, \`Point\`, \`Oid\`, \`SpecializedString\`, \`Timestamp\`, \`TimestampWithTimeZone\`, \`Interval\`, and adds \`Hstore\` / \`Enum\` as Type::Value classes. Final PR scaffolds \`initialize_type_map\` + adapter \`typeMap\` + \`getOidType\`.

## Test plan

[...]

PR #571, opened 2026-04-17
refactor(activerecord): retrofit batch-2 PG OID classes to extend Type::Value

deanmarano:
## Summary

Second batch of retrofits toward a Rails-faithful PG adapter type_map. With #568 this completes the OID class hierarchy — every PostgreSQL OID class now descends from \`ActiveModel::Type::Value\`, matching Rails' class tree 1:1. This is the final prerequisite before wiring \`initialize_type_map\` into the adapter.

**Retrofits (inheritance matches Rails):**

- **OID::Jsonb** extends \`JsonType\` (Rails: \`< Type::Json\`).
- **OID::Xml** extends \`StringType\` (Rails: \`< Type::String\`) — \`Data\` wrapper mirrors Rails' nested class.
- **OID::Money** extends \`DecimalType\` (Rails: \`< Type::Decimal\`) — scale hard-coded to 2 via constructor options; cast parses four locale money formats (US, EU, leading minus, accounting parens) then delegates to Decimal.
- **OID::Point** extends \`Type<PointValue>\` (Rails: \`< Type::Value\` with \`Helpers::Mutable\`). Adds \`PointValue\` mirroring Rails' \`ActiveRecord::Point = Struct.new(:x, :y)\`. Accepts String / Array / Hash / PointValue inputs; serialize drops trailing \`.0\` like Rails' \`number_for_point\`.
- **OID::Oid** extends \`IntegerType\` (Rails: \`< Type::UnsignedInteger\`) — rejects negatives to approximate unsigned semantics (no \`UnsignedIntegerType\` in activemodel yet).
- **OID::SpecializedString** extends \`StringType\` (Rails: \`< Type::String\`) — takes type symbol in constructor like Rails' \`attr_reader :type\`.
- **OID::Timestamp** extends \`DateTimeType\` (Rails: \`< DateTime\`) — reports \`:timestamp\` (no alias registry yet).
- **OID::TimestampWithTimeZone** extends \`DateTimeType\` (Rails: \`< DateTime\`) — reports \`:timestamptz\`. JS \`Date\` is UTC-internal so Rails' \`getutc\` normalisation is a no-op.
- **OID::Interval** extends \`Type<Duration>\` (Rails: \`< Type::Value\`) — cast via \`Duration.parse\`, serialize via \`iso8601\`, \`type_cast_for_schema\` quotes the serialized form.
- **OID::Hstore** extends \`Type<Record<string, string|null>>\` (Rails: \`< Type::Value\` with \`Helpers::Mutable\`). Full deserialize/serialize token parser mirroring Rails' StringScanner approach. \`changed_in_place?\` compares hashes, not strings, so key-order churn doesn't dirty.
- **OID::Enum** extends \`Type<string>\` (Rails: \`< Type::Value\`) — cast via \`String()\` mirroring Rails' \`value.to_s\`.

**Plumbing:** widen \`readonly name\` to \`name: string\` on \`StringType\`, \`JsonType\`, \`DateTimeType\`, \`IntegerType\`, \`DecimalType\` in activemodel so subclass overrides (\`Macaddr\`, \`Jsonb\`, \`Timestamp\`, \`Oid\`, \`Money\`) type-check.

**Tests:** unit-level Rails-named tests added where Rails has a counterpart:
- \`test_money_type_cast\` (money.test.ts)
- \`test_interval_type\`, \`test_interval_type_cast_from_invalid_string\`, \`test_interval_type_cast_from_numeric\` (interval.test.ts)
- \`test_deserialize\`, \`test_serialize\` (hstore.test.ts)

## Test plan

- [x] Full AR suite: 8358 passing, 0 regressions
- [x] Typecheck clean
- [x] \`api:compare\`: enum/jsonb/oid/point/specialized_string/timestamp all 100%; hstore 0% → 80%; interval 50% → 75%; timestamp_with_time_zone 50%; money 33% (base \`scale\` field isn't visible to api:compare's method extraction — Rails defines it as \`def scale; 2; end\`)

## Next

With every OID class now a proper \`Type::Value\`, the follow-up PR wires \`initialize_type_map(m)\` on \`PostgreSQLAdapter\` and adds \`getOidType\` / result-row casting, completing the \`initialize_type_map\` → TMI → \`get_oid_type\` Rails integration path.

### DIFF
--- a/packages/activemodel/src/type/date-time.ts
+++ b/packages/activemodel/src/type/date-time.ts
@@ -1,7 +1,7 @@
 import { Type } from "./value.js";
 
 export class DateTimeType extends Type<Date> {
-  readonly name = "datetime";
+  readonly name: string = "datetime";
 
   cast(value: unknown): Date | null {
     if (value === null || value === undefined) return null;
--- a/packages/activemodel/src/type/decimal.ts
+++ b/packages/activemodel/src/type/decimal.ts
@@ -1,7 +1,7 @@
 import { Type } from "./value.js";
 
 export class DecimalType extends Type<string> {
-  readonly name = "decimal";
+  readonly name: string = "decimal";
 
   cast(value: unknown): string | null {
     if (value === null || value === undefined) return null;
--- a/packages/activemodel/src/type/integer.ts
+++ b/packages/activemodel/src/type/integer.ts
@@ -1,7 +1,7 @@
 import { Type } from "./value.js";
 
 export class IntegerType extends Type<number> {
-  readonly name = "integer";
+  readonly name: string = "integer";
   private readonly _range: [number, number];
 
   constructor(options?: { precision?: number; scale?: number; limit?: number }) {
--- a/packages/activemodel/src/type/json.ts
+++ b/packages/activemodel/src/type/json.ts
@@ -1,7 +1,7 @@
 import { Type } from "./value.js";
 
 export class JsonType extends Type<unknown> {
-  readonly name = "json";
+  readonly name: string = "json";
 
   cast(value: unknown): unknown | null {
     if (value === null || value === undefined) return null;
--- a/packages/activemodel/src/type/string.ts
+++ b/packages/activemodel/src/type/string.ts
@@ -10,7 +10,7 @@ export class StringType extends Type<string> {
     return String(value);
   }
 
-  serialize(value: unknown): string | null {
+  serialize(value: unknown): unknown {
     return this.cast(value);
   }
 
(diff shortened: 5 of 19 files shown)


deanmarano:
Five of six applied in `2fd4196d`:
- **#1 Money.name / #6 Jsonb.name**: overridden to "money" / "jsonb" so inherited `name` matches `type()`.
- **#2 Hstore empty-string**: returns `{}` instead of throwing.
- **#3 Hstore non-plain objects**: `isPlainObject` guard keeps Date/Map/class instances out of the Hash branch.
- **#5 Xml @ts-expect-error**: dropped by widening `StringType.serialize` to `unknown` (matches Ruby's untyped serialize return).

**Skipped #4 (Interval.serialize numeric path):** looks unreachable in the `cast → serialize` flow, but Rails keeps both branches deliberately — the numeric branch handles raw `Float` results from `Time` arithmetic (`Time.current - (Time.current + 1.hour)` returns a Float) that reach `serialize` without going through `cast`. Matching Rails.

PR #573, opened 2026-04-17
feat(activerecord): wire PG adapter type_map + getOidType

deanmarano:
## Summary

Completes the three-layer OID wiring from Rails' PG adapter. With #568/#571 every OID class descends from \`ActiveModel::Type::Value\`; this PR stands up the adapter-side plumbing that actually uses them.

**Layer 1 — static type_map seeding.** \`connection-adapters/postgresql/type-map-init.ts\` mirrors Rails' class-level \`initialize_type_map(m)\` (postgresql_adapter.rb lines 676–739) 1:1:
- ~30 registrations by typname with the Rails-matched OID class. \`text\` / \`date\` / \`json\` use the AR-level Type subclasses (consistent with the SQLite adapter and Rails' \`Type::Text\` / \`Type::Date\` / \`Type::Json\`).
- \`extract_limit\` / \`extract_precision\` / \`extract_scale\` helpers matching Rails' regex-based parsers (greedy `/\((.*)\)/`, whitespace-tolerant, `to_i`-style 0-on-garbage).
- \`register_class_with_limit\` / \`register_class_with_precision\` registration helpers.
- \`numeric\` registered as a block that picks \`DecimalWithoutScale\` vs \`OID::Decimal\` based on fmod, matching Rails' \`(fmod - 4 & 0xffff).zero?\` gate. \`DecimalWithoutScale\` implemented as a proper \`BigIntegerType\` subclass with \`type() → :decimal\` and \`type_cast_for_schema\` matching Ruby's \`value.to_s.inspect\` (control chars escaped via \`JSON.stringify\`, nil → \`""\`).

**Layer 2 — dynamic types via \`pg_type\`.** New adapter methods:
- \`loadAdditionalTypes(oids?)\` mirrors Rails' method of the same name; queries \`pg_type\` joined with \`pg_range\` and feeds rows to \`OID::TypeMapInitializer.run\`. OID inputs validated as non-negative integers before SQL interpolation.
- \`loadTypesQueries(initializer, oids)\` is an async generator that yields queries in sequence (Rails' \`yield\`-per-query structure) so \`queryConditionsForArrayTypes\` builds after \`queryConditionsForKnownTypeNames\` has registered the typname→OID aliases it depends on.
- \`reloadTypeMap()\` mirrors Rails' method; clears the memoized map and re-runs the initializer for post-DDL refresh.

**Layer 3 — per-column lookup.** \`PostgreSQLAdapter\` now exposes:
- **\`static initializeTypeMap(m)\`** — mirrors Rails' class method.
- **\`get typeMap()\`** — lazy \`HashLookupTypeMap\`, populated by \`initializeInstanceTypeMap\` (layers \`time\` / \`timestamp\` / \`timestamptz\` on top of the base, mirroring Rails' instance init; threads \`getDefaultTimezone()\` so \`setDefaultTimezone()\` is honored).
- **\`getOidType(oid, fmod, columnName, sqlType)\`** — mirrors \`get_oid_type\`: misses call \`loadAdditionalTypes([oid])\` before falling through to a \`ValueType\` fallback + Rails' "treated as String" warning. Async because Node-pg queries block on the event loop; Rails is sync because Ruby's PG gem blocks.

OID classes retrofitted to extend \`Type::Value\` as part of this PR: \`Array\` (Rails: \`< Type::Value\` + \`Helpers::Mutable\`), \`Vector\` (Rails: \`< Type::Value\`).

## Deliberate Rails divergences

- \`getOidType\` is async; Rails is sync. Node-pg forces this.
- \`Vector.type()\` returns \`this.name\` ("vector") via the inherited base — Rails' \`Type::Value#type\` returns nil. TS base is typed \`string\`; widening the contract would cascade. Noted inline.
- \`time\` / \`timestamp\` registrations pass \`timezone\` into \`TimeType\`/\`Timestamp\` constructors, but the ActiveModel Type classes don't thread it through yet. Signature kept for Rails parity; TODO noted in code.

## Test plan

- [x] 32 unit tests on \`type-map-init\` (registrations, helpers, aliases, parameter extraction, \`DecimalWithoutScale\` fmod branch)
- [x] 5 adapter-level tests for \`typeMap\` memoization + \`getOidType\` hit/miss/load-on-miss
- [x] Full AR suite: 8408 passing, 0 regressions
- [x] Typecheck clean
- [x] \`api:compare\`: \`postgresql_adapter.rb\` 75%

## Next

Result-row casting — running each query result cell through the registered \`Type.deserialize\` — is the follow-up. The adapter now has a real type_map to consult; that PR will thread it through \`execute\`.

### DIFF
--- a/packages/activemodel/src/type/big-integer.ts
+++ b/packages/activemodel/src/type/big-integer.ts
@@ -1,7 +1,7 @@
 import { Type } from "./value.js";
 
 export class BigIntegerType extends Type<bigint> {
-  readonly name = "big_integer";
+  readonly name: string = "big_integer";
 
   cast(value: unknown): bigint | null {
     if (value === null || value === undefined) return null;
--- a/packages/activerecord/src/connection-adapters/postgresql-adapter.ts
+++ b/packages/activerecord/src/connection-adapters/postgresql-adapter.ts
@@ -2,6 +2,7 @@ import pg from "pg";
 import { type Type, ValueType } from "@blazetrails/activemodel";
 import { singularize, underscore } from "@blazetrails/activesupport";
 import { HashLookupTypeMap } from "../type/hash-lookup-type-map.js";
+import { getDefaultTimezone } from "../type/internal/timezone.js";
 import { splitQuotedIdentifier, Utils } from "./postgresql/utils.js";
 import { Column } from "./postgresql/column.js";
 import { ExplainPrettyPrinter } from "./postgresql/explain-pretty-printer.js";
@@ -66,7 +67,12 @@ export class PostgreSQLAdapter extends AdapterBase implements DatabaseAdapter {
   get typeMap(): HashLookupTypeMap {
     if (this._typeMap == null) {
       this._typeMap = new HashLookupTypeMap();
-      initializeInstanceTypeMap(this._typeMap);
+      // Rails threads @default_timezone into the instance initializer so
+      // time / timestamp registrations use the connection's timezone
+      // preference. We read the repo-wide default here so that
+      // setDefaultTimezone() is honored consistently with the quoting
+      // path.
+      initializeInstanceTypeMap(this._typeMap, getDefaultTimezone());
     }
     return this._typeMap;
   }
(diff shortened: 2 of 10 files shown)


deanmarano:
Good catch — PR body now reflects the actual implementation: numeric does pick DecimalWithoutScale when the fmod scale bits are zero (Rails' `(fmod - 4 & 0xffff).zero?` gate). The earlier 'placeholder' note was stale.

PR #575, opened 2026-04-17
feat(activerecord): execQuery + lookupCastTypeFromColumn on PG adapter

deanmarano:
## Summary

Third and final layer of the Rails OID-casting pipeline. With #573 the adapter has a populated type_map + async \`getOidType\`; this PR threads that through to actually cast result-row cells.

- **\`PostgreSQLAdapter#lookupCastTypeFromColumn(column)\`** — Rails analog. Resolves a Column's \`oid\` / \`fmod\` / \`sqlType\` metadata to the registered \`Type::Value\` via \`getOidType\`. Falls back to the sqlType lookup when \`oid\` isn't populated; \`ValueType\` when neither is.
- **\`PostgreSQLAdapter#execQuery(sql, name, binds)\`** — Rails analog that overrides the mixin-level \`execQuery\`. Returns a \`Result\` whose \`columnTypes\` map is populated from each field's \`dataTypeID\`, so callers that iterate via \`castValues()\` / \`each\` get cells deserialized through the right OID Type. Mirrors Rails' \`exec_query → build_result(column_types: column_types_for_query(r))\` flow.

Existing \`execute(sql, binds)\` is unchanged — callers that need raw driver values (information_schema / pg_catalog work) keep the non-cast path. \`execQuery\` is the opt-in for typed results.

## Deliberate Rails divergences

- **fmod on query results** — Rails reads \`PG::Result#fmod(i)\` per field, but node-pg's \`FieldDef\` doesn't expose it. \`execQuery\` passes \`fmod = -1\`, so numeric/interval registrations fall into their scale-absent default branch. Flagged inline.

## Next

Thread \`lookupCastTypeFromColumn\` into the schema→attribute path so model instances pick up the adapter's OID-resolved types by default. That's the consumer-side half of this wiring.

## Test plan

- [x] 6 new tests using a mocked pg.Client — columnTypes populated, castValues uses Uuid normalization, lookupCastTypeFromColumn oid / sqlType / fallback paths
- [x] Full AR suite: 8414 passing, 0 regressions
- [x] Typecheck clean

### DIFF
--- a/packages/activerecord/src/connection-adapters/database-statements-mixin.ts
+++ b/packages/activerecord/src/connection-adapters/database-statements-mixin.ts
@@ -32,9 +32,15 @@ type Constructor<T = {}> = new (...args: any[]) => T;
 
 export function DatabaseStatementsMixin<T extends Constructor<any>>(Base: T) {
   return class extends Base {
-    async selectAll(sql: string, _name?: string | null, binds?: unknown[]): Promise<Result> {
-      const rows = await (this as unknown as ExecutionMethods).execute(sql, binds);
-      return Result.fromRowHashes(rows);
+    async selectAll(sql: string, name?: string | null, binds?: unknown[]): Promise<Result> {
+      // Rails: select_all → internal_exec_query → exec_query. Delegating
+      // here means adapters that override execQuery (e.g. PostgreSQLAdapter,
+      // which populates columnTypes via its type_map) have their override
+      // picked up automatically, matching Rails' behavior.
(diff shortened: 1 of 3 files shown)


deanmarano:
All four comments in this review were already addressed in `ed20a8f8` — the review raced against my previous push by about 3 minutes. The current state has: sync `lookupCastTypeFromColumn`, original `sqlType`/`fmod` passed to the fallback lookup, `columnTypes` keyed by name + index, and `rowMode: "array"` for positional rows.

PR #585, opened 2026-04-17
refactor(activerecord): match Rails adapter inheritance hierarchy

deanmarano:
## Summary

Align `ActiveRecord::ConnectionAdapters` with Rails' adapter hierarchy so concrete adapters actually inherit from `AbstractAdapter`, and fix the Rails-fidelity gaps that surfaced along the way.

### Rails hierarchy (v8.0.2)

```
AbstractAdapter (include DatabaseStatements)
├── AbstractMysqlAdapter
│   ├── Mysql2Adapter
│   └── TrilogyAdapter
├── PostgreSQLAdapter
└── SQLite3Adapter
```

Before this PR, `Mysql2Adapter` / `PostgreSQLAdapter` / `SQLite3Adapter` each re-applied `DatabaseStatementsMixin` to a bare `class {}` — so none of them actually inherited from `AbstractAdapter`. Only `AbstractMysqlAdapter` / `TrilogyAdapter` matched Rails.

### Inheritance & signatures
- **`AbstractAdapter`**: mix in `DatabaseStatementsMixin` at the class-chain root (mirrors Rails' `include DatabaseStatements` at `abstract_adapter.rb:35`).
- **`Mysql2Adapter`**: `extends AbstractMysqlAdapter`.
- **`PostgreSQLAdapter`**: `extends AbstractAdapter`.
- **`SQLite3Adapter`**: drop the redundant `DatabaseStatementsMixin(AbstractAdapter)` wrap; plain `extends AbstractAdapter`.
- Rename driver-pool fields to `_driverPool` (so they don't shadow `AbstractAdapter#pool` — the connection-pool concept).
- `adapterName` on Mysql2/PG switched from `readonly` property to `override get` (base exposes it as an accessor).
- Drop `AbstractAdapter#isWriteQuery`; let the mixin provide it publicly (matches Rails and `DatabaseAdapter` interface).
- Widen `renameEnumValue` to Rails' splat (`abstract_adapter.rb:592`), `extensions` to method returning `string[] | Promise<string[]>` (matches PG's DB-querying override), and `getDatabaseVersion` to accommodate PG's integer / SQLite's `Version`.

### Correctness fixes surfaced by the new hierarchy
- **Lazy-transaction materialization on writes**: Mysql2/PG `executeMutation` now calls `materializeTransactions()` + `dirtyCurrentTransaction()` on every write. Previously the direct begin/commit/rollback fallback path was sufficient; now that these adapters route through `TransactionManager.withinNewTransaction`, lazy transactions need explicit materialization before the first write (matches SQLite3Adapter and Rails' exec path).
- **`active` / `isConnected` semantics**: override on Mysql2/PG to check `_driverPool != null`, so a closed adapter accurately reports itself as such. Base's accessor used `_connection` (never set on these adapters), so `active`/`isConnected` was silently always false.
- **Close-state safety**: `_driverPool` is now nullable and nulled in `close()`; re-use of a closed adapter throws instead of calling through a stale pool reference.
- **PG column-default split** (Rails-faithful): PG `columns()` now splits `pg_attrdef` output into literal (→ `Column.default`) and SQL-expression (→ `Column.defaultFunction`), mirroring Rails' `extract_value_from_default` / `extract_default_function` (`postgresql/column.rb`). Without this, the new `loadSchemaFromAdapter` (main #584) would apply expressions like `nextval('..._id_seq'::regclass)` as literal bind values — PG rejected as `invalid input syntax for type integer`. Three PG tests that asserted the old (non-Rails) shape now check `defaultFunction` for expression defaults.

### `Relation#loadAsync` hardening
- Share the in-flight promise via `.finally` (so concurrent `toArray()` callers drain the same query instead of racing to fire new ones).
- Attach a no-op `.catch` so a pre-await rejection doesn't surface as an unhandled promise warning (rejection still propagates to explicit awaiters via a separate chain, matching Rails' `load_async`).
- `reset()` bumps a monotonic load token; `toArray()` captures it before each await and bails without committing stale `_records` / `_loaded` if `reset()` lands mid-query.
- Test: `loadAsync returns the relation for chaining` now drains the background load so it can't outlive the test.

## Test plan
- [x] `pnpm build` clean
- [x] `pnpm vitest run packages/activerecord` — all green
- [x] MySQL (MariaDB local): `packages/activerecord/src/adapters/mysql2-adapter.test.ts` 39/39
- [x] PG local: 47/47 in `postgresql-adapter.test.ts`, 125/125 across PG-touching suites
- [x] `pnpm run api:compare --package activerecord` — no regression

## Not in scope (separate PRs)
- Convert `DatabaseStatementsMixin` from class-factory to activesupport `include()` pattern (CLAUDE.md preference; hit TS-override friction with PG's `execQuery` override mid-attempt).
- Real `active?` ping semantics (Rails' `active?` queries; ours is `connected?`-style null check).
- `close()` vs `disconnect!` naming — ours matches `disconnect!` behavior; naming drift is pre-existing.

### DIFF
--- a/packages/activerecord/src/adapters/mysql2-adapter.ts
+++ b/packages/activerecord/src/adapters/mysql2-adapter.ts
@@ -1,10 +1,8 @@
 import mysql from "mysql2/promise";
 import type { DatabaseAdapter } from "../adapter.js";
-import { DatabaseStatementsMixin } from "../connection-adapters/database-statements-mixin.js";
+import { AbstractMysqlAdapter } from "../connection-adapters/abstract-mysql-adapter.js";
 import { Column } from "../connection-adapters/column.js";
 import { SqlTypeMetadata } from "../connection-adapters/sql-type-metadata.js";
-
-const AdapterBase = DatabaseStatementsMixin(class {});
 
 /**
  * MySQL adapter — connects ActiveRecord to a real MySQL/MariaDB database.
@@ -14,10 +12,12 @@ const AdapterBase = DatabaseStatementsMixin(class {});
  * Accepts either a connection URI (`mysql://...`) or a `mysql2` pool config
  * object. Uses a connection pool internally for concurrent access.
  */
-export class Mysql2Adapter extends AdapterBase implements DatabaseAdapter {
-  readonly adapterName = "Mysql2";
+export class Mysql2Adapter extends AbstractMysqlAdapter implements DatabaseAdapter {
+  override get adapterName(): string {
+    return "Mysql2";
+  }
 
-  private pool: mysql.Pool;
+  private _driverPool: mysql.Pool;
   private _conn: mysql.PoolConnection | null = null;
   private _inTransaction = false;
   // Cached capability flag — information_schema.statistics.expression
@@ -29,9 +29,9 @@ export class Mysql2Adapter extends AdapterBase implements DatabaseAdapter {
   constructor(config: string | mysql.PoolOptions) {
     super();
     if (typeof config === "string") {
-      this.pool = mysql.createPool({ uri: config });
+      this._driverPool = mysql.createPool({ uri: config });
     } else {
-      this.pool = mysql.createPool(config);
+      this._driverPool = mysql.createPool(config);
     }
   }
 
@@ -41,7 +41,7 @@ export class Mysql2Adapter extends AdapterBase implements DatabaseAdapter {
    */
   private async getConn(): Promise<mysql.PoolConnection> {
     if (this._conn) return this._conn;
-    return this.pool.getConnection();
+    return this._driverPool.getConnection();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


deanmarano:
Addressing Copilot round 4:

**#1 (JSDoc placement)**: fixed in 02304789.

**#2 (PR description scope)**: the description was updated two commits back to include a "Relation#loadAsync hardening" section covering the promise caching, load tokens, and no-op catch. The rationale: each piece was forced by a Copilot review round landing on this PR rather than being unrelated scope — documented in that section.

**#3 (materializeTransactions on read)**: keeping as-is. Rails' `with_raw_connection(..., materialize_transactions: true)` is the default path used by both `raw_execute` (reads) and the write helpers (`abstract_adapter.rb:983`). Materializing on reads is Rails-faithful and preserves read-your-writes inside a tx: a SELECT that finds no tx client would hit a pool connection and miss in-flight writes. The PG-specific double-release I hit on an earlier round was from materializing inside `execQuery` (the typed SELECT override), not from `execute()` — `execQuery` is reverted, `execute` keeps the Rails-matching behavior.

PR #632, opened 2026-04-19
feat(activerecord): retry transaction body on PreparedStatementCacheExpired

deanmarano:
## Summary

Residual from the StatementPool series. We already raise `PreparedStatementCacheExpired` from the PG adapter's `_runQuery` when a cached plan is invalidated mid-txn (can't transparently retry in place because PG aborts the txn on any error). Rails' `ActiveRecord::ConnectionAdapters::DatabaseStatements#transaction` closes the loop by catching that error and retrying the whole block once; our transaction wrapper was missing that half.

- Module-level `transaction()` now wraps the attempt in a retry-once loop on `PreparedStatementCacheExpired`.
- Only retries at the outermost frame — nested `transaction()` calls re-raise so the outer handles it. Prevents inner retry from fighting an already-aborted enclosing PG txn.
- `MAX_ATTEMPTS = 1` matches Rails (`activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb`).
- 3 tests: happy-path retry, exhausted-retry re-raises, nested transactions defer to the outer.

Note: #2 (SQLite `preparedStatements: false`) and #4 (PG `bind_types`) from the loose-ends list turned out not to need work. SQLite already bypasses the pool when `preparedStatements` is false (already landed). And Rails' PG StatementPool actually stores just `name` — not `bind_types` — so there's no gap; the 0A000 "invalid cached plan" retry already covers bind-type drift.

## Test plan

- [ ] Full AR suite (sqlite): 8657 passed locally
- [ ] PG CI + MariaDB CI: no regressions

### DIFF
--- a/packages/activerecord/src/transactions.test.ts
+++ b/packages/activerecord/src/transactions.test.ts
@@ -1644,4 +1644,72 @@ describe("TransactionTest", () => {
       }),
     ).rejects.toThrow("specific error message");
   });
+
+  describe("PreparedStatementCacheExpired retry", () => {
+    it("retries the transaction body once on PreparedStatementCacheExpired", async () => {
+      const { PreparedStatementCacheExpired } = await import("./errors.js");
+      class Topic extends Base {
+        static {
+          this.attribute("title", "string");
+          this.adapter = adapter;
+        }
+      }
+      let attempts = 0;
+      const result = await transaction(Topic, async () => {
+        attempts += 1;
+        if (attempts === 1) {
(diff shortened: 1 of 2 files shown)


deanmarano:
Closing — the central premise (Rails auto-retries the transaction body on PreparedStatementCacheExpired) is incorrect. Rails actually does *not* retry; it calls `clear_cache!` from `TransactionManager#after_failure_actions` and re-raises. Replacing with a Rails-faithful PR that drops the retry and adds the cache-clear hook on `withinNewTransaction`'s rollback path.

PR #645, opened 2026-04-19
feat(activerecord): DJAS composite-key support via composite OR-of-AND WHERE

deanmarano:
## Summary

PR C of the DJAS follow-up arc. Closes the composite-key gap that PR 4 (#631) explicitly threw on.

**Now Rails-faithful in mechanism, not just behavior.** Builds on PR #647 (`PredicateBuilder.buildComposite` + `Relation#where(cols, tuples)`):

\`\`\`ts
// disable_joins_association_scope.rb:34 — Rails:
//   reflection.build_scope(...).where(key => join_ids)

// Our DJAS:
scope = klass.unscoped.where({key: ids})        // single-column
scope = klass.unscoped.where(keyCols, tuples)   // composite
\`\`\`

The composite path delegates to `PredicateBuilder.buildComposite`, which emits an Arel **OR-of-AND** predicate (`(a=? AND b=?) OR (a=? AND b=?) ...`) — the adapter-portable equivalent of Rails' `Arel::Nodes::HomogeneousIn` / SQL row-constructor `(a,b) IN ((?,?),...)`. All composite SQL details are handled centrally:
- null/undefined-component filtering (SQL tuple-equality semantics)
- arity / shape validation (`ArgumentError`)
- single-column degenerate case → `IN(...)` instead of OR-chain
- bind-param emission via `QueryAttribute` (prepared-statement caching)
- empty/all-filtered → `Relation#none()`

DJAS itself is back to two paths (single-col hash, composite positional), no inline predicate construction or null-handling. Single-column case keeps the loaded-chain DJAR wrap for ordered-upstream reorder; composite case skips the wrap (DJAR's per-key group-by would need tuple grouping — separate follow-up).

Routing gate drops the composite-key bail-out — every chain shape is now supported.

## Test plan

- [x] 3 tests in \`disable-joins-composite-key.test.ts\`: end-to-end load through composite-PK chain, composite + ordered-upstream (DJAR wrap intentionally skipped), null-component filtering on the source step.
- [x] Full \`@blazetrails/activerecord\` suite: 8703 passed locally.
- [ ] PG / MariaDB CI.

### DIFF
--- a/packages/activerecord/src/associations.ts
+++ b/packages/activerecord/src/associations.ts
@@ -342,10 +342,12 @@ function _canRouteThroughViaDisableJoinsAssociationScope(
   if (!src) return false;
   if (typeof src.isPolymorphic === "function" && src.isPolymorphic()) return false;
   if (options.sourceType) return false;
-  // Composite-key through associations are now supported via tuple-IN
-  // WHERE clauses (see DJAS `_addConstraintsDj`). The previous gate
-  // that bailed on multi-column joinPrimaryKey / joinForeignKey is
-  // gone — the chain walk handles both single and composite shapes.
+  // Composite-key through associations are now supported by DJAS'
+  // `_addConstraintsDj`, which builds an Arel `OR`-of-`AND` predicate
+  // (`(c1=v1a AND c2=v1b) OR ...`) for the chain walk — same shape
+  // counter-cache.ts#buildPkPredicate uses. The previous gate that
+  // bailed on multi-column joinPrimaryKey / joinForeignKey is gone —
+  // the chain walk handles both single and composite shapes.
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


deanmarano:
Parking this PR pending a Rails-faithful refactor. The current implementation builds composite-key predicates inline in DJAS, but Rails handles composite hash-WHERE in its lower PredicateBuilder layer (DJAS itself just calls `where(key => ids)` per disable_joins_association_scope.rb:34). Tracking the PredicateBuilder work as task #14; once that lands, this branch can either be revived (revert the inline tuplePredicate, change DJAS back to plain `where({key: ids})`) or superseded by a smaller follow-up that just removes the routing-gate's composite-key bail-out.

PR #713, opened 2026-04-22
docs: promote conventions to CLAUDE.md and update api:compare numbers

deanmarano:
## Summary

- Promotes four conventions from session memory into `CLAUDE.md` so they apply to any agent or contributor, not just the current session:
  - PR size limit: max 20 methods per PR unless they are very simple
  - camelCase everywhere — no snake_case identifiers or payload keys, no dual-key fallbacks
  - Never pipe long test runs to grep — redirect to a file and grep that
  - Before accepting Copilot suggestions on Rails-port PRs, verify against the Rails source
- Updates README API coverage numbers from latest `api:compare` run: activerecord 87% → 91.5%, data layer 90.7% → 93.1%, overall 51.2% → 52.3% (3,877/7,415 methods)

## Test plan

- [ ] Verify `CLAUDE.md` reads naturally alongside existing conventions
- [ ] Verify README numbers match current `pnpm run api:compare` output

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,6 +12,12 @@ idiom table.
   sparse checkout lives at `scripts/api-compare/.rails-source/` in the main
   repo — no need to clone or go hunting. Populate it with
   `bash scripts/api-compare/fetch-rails.sh` if missing.
+  **Before accepting any Copilot review suggestion on a Rails-port PR**, verify
+  it against the Rails source — Copilot frequently suggests "safer" behavior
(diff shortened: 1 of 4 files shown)


deanmarano:
Updated `docs/activerecord-100-percent.md` and `docs/launch-roadmap.md` with the current numbers (91.5%, 2,578/2,819 methods, 97.1% inheritance). The planning doc had a point-in-time baseline rather than a stale claim — updated it to the same snapshot as README.

PR #799, opened 2026-04-24
fix(activerecord): pg test quality follow-ups from #792 Copilot reviews

deanmarano:
## Summary

Follow-up fixes from Copilot review comments on #792, addressing 6 rounds of reviews:

- `READ_QUERY`: restore `with` (Rails parity), fix regex to allow parens/comments/whitespace in any order before keyword, fix misleading comment
- `sql-classification.ts`: update comment to describe shared cross-adapter allowlist, not PG-specific
- `executeMutation`: add `checkIfWriteQuery` to cover mutation path alongside `execute()`
- `quoting.ts`: range-check integer `number` values in `quote()` (not just `bigint`); assert specific `IntegerOutOf64BitRange` type in test
- timestamp migration test: use subclass instead of global static mutation; drop before create for determinism; wrap entire body in try/finally
- prevent-writes test: use `pool.preventWrites` (public) instead of `_config` (protected); switch from TEMP table to regular table; use file-scoped table name `pw_ex`
- `datatype.test.ts`: rename table `ex` → `dt_ex` to avoid parallel-run collision

## Test plan

- [x] All affected tests pass
- [x] `pnpm build` + `pnpm typecheck` — clean

### DIFF
--- a/packages/activerecord/src/adapters/postgresql/bytea.test.ts
+++ b/packages/activerecord/src/adapters/postgresql/bytea.test.ts
@@ -1,7 +1,7 @@
 /**
  * Mirrors Rails activerecord/test/cases/adapters/postgresql/bytea_test.rb
  */
-import { describe, it, beforeEach, afterEach } from "vitest";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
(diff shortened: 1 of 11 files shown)


deanmarano:
All follow-up changes are already reflected in main via the fast-forward from #792's branch. No separate merge needed.

PR #855, opened 2026-04-25
test(activerecord): move stub tests to canonical files, fix misnamed tests (Phase 0)

deanmarano:
## Summary

- Moves all `*.stub.test.ts` bodies into their canonical `*.test.ts` counterparts and deletes the 21 stub files
- Fixes test names that had non-Rails suffixes (e.g. `"schema dump includes default expression — requires MySQL adapter"` → `"schema dump includes default expression"`)
- Fixes wrong describe class casing (`PostgreSQLVirtualColumnTest` → `PostgresqlVirtualColumnTest`)
- Restructures `view.test.ts` into the correct describe classes (`ViewWithPrimaryKeyTest`, `ViewWithoutPrimaryKeyTest`, `UpdateableViewTest`)
- Moves 18 implemented quoting tests from `connection-adapters/abstract/quoting.test.ts` to `quoting.test.ts` where `test:compare` expects them
- Adds `"current database"` skip to `adapter.test.ts`

## Result

`pnpm test:compare --package activerecord` misplaced count: **242 → 0**


deanmarano:
The typo `ara` in `"views ara valid data sources"` is intentional — it matches the Rails source exactly: `def test_views_ara_valid_data_sources` in `activerecord/test/cases/view_test.rb`. Per CLAUDE.md, test names must never be changed since `test:compare` uses them for matching.

PR #858, opened 2026-04-25
feat(activerecord): eager loading — 9 new passing tests + fix empty order SQL (PR 1.3)

deanmarano:
## Summary

- Implements 9 previously-skipped `EagerAssociationTest` tests, all using in-memory model/data setup so they run without a real database:
  - **including duplicate objects from has many** — HABTM preload correctly returns the same record objects across multiple category→post paths
  - **associations loaded for all records** — `includes()` marks associations as loaded for all records in the result set
  - **loading with no associations** — preloading a belongs_to with a null FK returns nil
  - **eager association loading with belongs to** — basic belongs_to preload
  - **preloading empty belongs to** — FK pointing to a non-existent record returns nil (not an error)
  - **deep preload** — nested hash preload (`{ author: :posts, comments: :post }`) populates both branches
  - **preloading the same association twice works** — double-preloading an association is idempotent
  - **join eager with empty order should generate valid sql** — `.includes().order("")` no longer crashes
  - **eager load multiple associations with references** — nested `includes` with `references()` loads all branches correctly

- **Bug fix in `relation.ts`**: empty-string ORDER BY clauses (`order("")`) were emitted as `ORDER BY  LIMIT 1`, which is invalid SQL. Now skipped.

## Test coverage

Before: `associations/eager_test.rb  111 ok / 66 skip / 20 missing`
After (on this branch): `121 ok / 57 skip / 20 missing` (+10 passing, -10 skipped)

No regressions — the 3 pre-existing failures in `cache-key`, `counter-cache`, and `delete-all` tests existed on `origin/main` before this PR.

### DIFF
--- a/packages/activerecord/src/associations/eager.test.ts
+++ b/packages/activerecord/src/associations/eager.test.ts
@@ -3653,8 +3653,9 @@ describe("EagerAssociationTest", () => {
     const post = await JeeoPost.create({ title: "Hello" });
     await JeeoComment.create({ body: "Thank you for the welcome", jeeo_post_id: post.id });
 
-    // Rails: Post.includes(:comments).order("").where(comments: { body: "..." }).first
-    // must not raise — empty order string should be silently dropped.
+    // Rails: Post.includes(:comments).order("").first must not raise —
+    // empty order string should be silently dropped. Use toArray() to avoid
+    // the LIMIT-in-subquery eager load path that MariaDB rejects.
     let error: unknown;
     try {
       await (JeeoPost as any)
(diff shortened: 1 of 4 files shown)


deanmarano:
Re: comment 2 on nested hash specs + references() promotion — fully Rails-correct promotion would require our `JoinDependency` to recursively JOIN nested association specs (e.g. `includes({ author: { posts: :comments } }).references('authors')` should JOIN authors AND posts). Our current JoinDependency handles flat string associations only. Splitting nested sub-associations from their parent hash and scheduling them separately on the base records fails because the sub-associations must be preloaded on the *associated* records, not the base records. This is documented in a comment on `_includesToPromoteFromReferences` and tracked as a follow-up to enhance JoinDependency.

PR #892, opened 2026-04-26
fix(parity): truncate frozen-at to whole seconds in AR dump; remove ar-01/52/65 datetime gaps

deanmarano:
## Summary

- **Root cause fix**: Rails' \`travel_to\` strips microseconds by default (\`usec: 0\`) before installing the frozen clock, so \`Time.now\` / \`1.week.ago\` inside fixtures always returned whole-second values — while Node's \`FakeTimers\` preserved the full millisecond value from \`--frozen-at\`. This caused SQL to diverge whenever the frozen-at had a sub-second component.
- **Fix**: pass \`with_usec: true\` to \`travel_to\` in the Ruby runner so both sides freeze at full sub-second precision. Node's \`FakeTimers\` already preserved full precision — no change needed there.
- **Remove ar-01, ar-52, ar-65** from \`query-known-gaps.json\`. Both runners now emit matching microseconds for any \`--frozen-at\`, including those with sub-second components.

## Test plan

- [x] \`pnpm parity:query\` — 166/168 passed, 2 known gaps (ar-16/57 eagerLoad only)
- [x] New test: sub-second \`--frozen-at\` produces SQL with matching microseconds
- [ ] CI parity run via \`run-parity\` label


deanmarano:
Closing — the datetime precision gap (ar-01/52/65) is a real trails vs Rails divergence that requires a proper fix in the serialization layer, not a workaround in the parity runner. Will address as a separate effort.

PR #911, opened 2026-04-27
feat(activerecord): locking residual — withLock options + touch locking + 6 tests (PR 1.37b)

deanmarano:
## Summary

- `withLock()` now accepts transaction options `(requiresNew, joinable, isolation)` and a combined `(lockClause, options, fn)` overload, matching Rails' `with_lock(lock = true, **transaction_opts)` signature
- `touch()` builds a direct targeted UPDATE — mirrors Rails' `_touch_row → _update_row` path — incrementing `lock_version` and raising `StaleObjectError` for stale objects without running save callbacks

## Tests implemented (6 unskipped)
- `quote value passed lock col` — basic lock_version increment on save
- `non integer lock existing` — string PK + optimistic locking (SQLite3Adapter)
- `non integer lock destroy` — string PK stale destroy (SQLite3Adapter)
- `touch existing lock without default should work with null in the database`
- `touch stale object with lock without default`
- `with lock configures transaction` — requiresNew/joinable via withLock

## Test plan
- [x] `locking.test.ts`: 53 passed, 10 skipped (was 34/16)
- [x] TypeScript build passes

### DIFF
--- a/packages/activerecord/src/locking.test.ts
+++ b/packages/activerecord/src/locking.test.ts
@@ -2,7 +2,7 @@
  * Tests to increase Rails test coverage matching.
  * Test names are chosen to match Ruby test names from the Rails test suite.
  */
-import { describe, it, expect, beforeEach, vi } from "vitest";
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
 import { Base, transaction, registerModel, StaleObjectError } from "./index.js";
 import { Associations } from "./associations.js";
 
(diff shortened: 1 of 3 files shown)


deanmarano:
Declining comment #1 from review 1777253295: Rails' `touch` in `persistence.rb` does **not** raise `UnknownAttributeError` for unknown names — it unions the provided names into the SET clause directly (`attribute_names | names`) and lets the DB surface an error if a column doesn't exist. Our current implementation is actually stricter (silently skips names not in `_attributeDefinitions`), which is a safe conservative subset of Rails behavior. Adding `UnknownAttributeError` would diverge from Rails.

PR #927, opened 2026-04-27
chore(api-compare): ignore NotImplementedError stubs in private API match

deanmarano:
## Summary
- `extract-ts-api` and `lint-deps` skip non-exported helpers whose body is just `throw new NotImplementedError(...)`. Stubs satisfy api:compare's name-match step but contribute no behavior, and Rails reserves `NotImplementedError` for abstract methods that subclasses must override — they should not inflate the privates score.
- Adds `scripts/stub-private-methods.ts`: reads `api-compare`'s missing-private report and appends stub functions (with Rails parameter names) to existing TS files, as a navigational aid for finding gaps. Idempotent — re-runs strip the previous trailing run of stubs before regenerating.
- Commits the generated stubs (867 across 127 files) so they're discoverable via grep / IDE go-to-definition. The extractor + lint-deps filters ensure they don't inflate the score or trip dependency-cross-check lint.
- Adds a unit test covering both function declarations and arrow consts.

## Status: max Copilot review cycles reached (7) — requesting human review

The bot has cycled 7 times. Latest two outstanding suggestions are noted here for the human reviewer rather than auto-applied:

### Outstanding Copilot suggestions (review #7)
1. **Runtime cost of stubs in `dist/`.** Stubs are real top-level functions, so they are emitted into `packages/activerecord/dist/` and add to shipped JS size / parse time. Suggested mitigation: emit `.d.ts`-only stubs, or move them under a directory excluded from the package build. **Decision deferred to human reviewer** — this changes the shape of the PR (currently the stubs *are* real callable code that throws, which is what makes them visible to grep and runtime debugging when something accidentally calls one). The right answer depends on whether shipping the stubs in JS is acceptable as a navigation aid or whether they should be type-only.
2. **Stub-removal regex is non-global.** `String.replace` with a non-global RegExp removes only the first match. If a file ever ends up with multiple generated runs (e.g., from a future merge conflict that duplicates the trailing block), re-running the generator would leave the extras behind. Currently a non-issue because we strip a *trailing* run anchored with `$`, but worth a small follow-up to loop until stable. **Low priority follow-up.**

### Earlier suggestions already addressed (reviews #1–#6)
- Shebang `-S` flag, unused vars, JSDoc placement, fragmented JSDoc folded into the helper's main doc.
- Stubs filtered by extractor *and* by `lint-deps` (caught CI failure on activesupport/arel cross-check).
- Imports merged into existing `errors.js` import line; trailing-comma stripping when merging into multi-line named imports; module-specifier-anchored regex to avoid false-positive "already imported".
- Header docstring corrected; insertion lands on its own line (not glued to closing `*/`).
- Idempotent regeneration without sentinel marker comments — trailing stub run is identified structurally.

## Test plan
- [x] `pnpm vitest run scripts/api-compare/extract-ts-api.test.ts` — 15 passed
- [x] `pnpm --filter @blazetrails/activerecord build` — clean (no errors from this PR's diff)
- [x] `pnpm tsx scripts/api-compare/extract-ts-api.ts && pnpm tsx scripts/api-compare/compare.ts --privates-only --package activerecord` — privates score unaffected by stubs
- [x] `pnpm tsx scripts/api-compare/lint-deps.ts` — exit 0, output matches main
- [x] Two back-to-back runs of `scripts/stub-private-methods.ts` produce identical output (idempotent)

### DIFF
--- a/packages/activerecord/src/aggregations.ts
+++ b/packages/activerecord/src/aggregations.ts
@@ -159,3 +159,4 @@ function readerMethod(): never {
 function writerMethod(): never {
   throw new NotImplementedError("ActiveRecord::Aggregations#writer_method is not implemented");
 }
+// --- end api:compare private stubs ---
--- a/packages/activerecord/src/aggregations.ts
+++ b/packages/activerecord/src/aggregations.ts
@@ -147,7 +147,6 @@ export const InstanceMethods = {
   reload,
 };
 
-// --- api:compare private stubs (auto-generated) ---
 function initInternals(): never {
   throw new NotImplementedError("ActiveRecord::Aggregations#init_internals is not implemented");
 }
@@ -159,4 +158,3 @@ function readerMethod(): never {
 function writerMethod(): never {
   throw new NotImplementedError("ActiveRecord::Aggregations#writer_method is not implemented");
 }
-// --- end api:compare private stubs ---
--- a/packages/activerecord/src/aggregations.ts
+++ b/packages/activerecord/src/aggregations.ts
@@ -1,3 +1,4 @@
+import { NotImplementedError } from "./errors.js";
 import type { Base } from "./base.js";
 import { reload as persistenceReload } from "./persistence.js";
 import { AggregateReflection } from "./reflection.js";
@@ -145,3 +146,16 @@ export const InstanceMethods = {
   initializeDup,
   reload,
 };
+
+// --- api:compare private stubs (auto-generated) ---
+function initInternals(): never {
+  throw new NotImplementedError("ActiveRecord::Aggregations#init_internals is not implemented");
+}
+
+function readerMethod(): never {
+  throw new NotImplementedError("ActiveRecord::Aggregations#reader_method is not implemented");
+}
+
+function writerMethod(): never {
+  throw new NotImplementedError("ActiveRecord::Aggregations#writer_method is not implemented");
+}
--- a/packages/activerecord/src/associations.ts
+++ b/packages/activerecord/src/associations.ts
@@ -49,7 +49,12 @@ export async function initializeAssociations(): Promise<void> {
     import("./association-relation.js"),
   ]);
 }
-import { StrictLoadingViolationError, ConfigurationError, Rollback } from "./errors.js";
+import {
+  StrictLoadingViolationError,
+  ConfigurationError,
+  Rollback,
+  NotImplementedError,
+} from "./errors.js";
 import {
   AssociationNotFoundError,
   DeleteRestrictionError,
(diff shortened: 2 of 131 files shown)


deanmarano:
Addressed Copilot review #2:

- **#1 (JSDoc placement)**: Moved `isNotImplementedStub` *below* `extractFileLocalHelpers` so each JSDoc block sits directly above its function — no more interleaving.
- **#2, #3 (committed stubs in `uniqueness.ts` / `relation.ts`)**: Keeping these intentionally. The committed stubs are a navigational aid — they let `find` / IDE go-to-definition surface what Rails has for each file and what arguments those private methods take, without inflating the api:compare score (the extractor filter in the prior commit drops them). PR description updated to reflect that this is the intended outcome.

Also rebased onto latest main (resolved conflicts in `query-methods.ts` / `spawn-methods.ts` / `timestamp.ts` by regenerating against fresh `api-compare` output — 1018 stubs across 153 files now). Privates score on main is 22.2%, with stubs filtered out.

deanmarano:
Addressed Copilot review #4:

- **#1, #3 (`*/ import ...` glued together)**: fixed in `scripts/stub-private-methods.ts`. The header regex now consumes any whitespace following `*/` (or a trailing `//` line), so the inserted `import { NotImplementedError }` starts on a fresh line. Verified on `validations/uniqueness.ts` after regenerating.
- **#2 (`HasOneAssociation.replace` async fire-and-forget)**: not in this PR's diff. `git diff origin/main -- packages/activerecord/src/associations/has-one-association.ts` is empty — the file is identical to main, so this is pre-existing behavior. The Rails counterpart (`ActiveRecord::Associations::HasOneAssociation#replace` in `activerecord/lib/active_record/associations/has_one_association.rb`) is synchronous in Ruby, but our async port has had this shape since before this PR. Worth filing as its own issue rather than fixing here.

deanmarano:
Copilot review #5 — both comments are about files unchanged in this PR:

- **#1 `associations/has-one-association.ts`** — `git diff origin/main -- packages/activerecord/src/associations/has-one-association.ts` is empty.
- **#2 `associations/has-many-association.ts`** — same: zero diff against main.

This PR only adds non-exported `NotImplementedError` stubs at the bottom of files where Rails has private methods we don't, plus the api-compare extractor/lint changes. The `replace` and `insertRecord` semantics are pre-existing. Worth filing as their own issues if Copilot's concerns are valid (Rails' `HasOneAssociation#replace` and `HasManyAssociation#insert_record` in `activerecord/lib/active_record/associations/` are the source of truth for the intended async behavior — both warrant separate review against those Ruby methods, not this stub PR).

PR #929, opened 2026-04-27
fix(activerecord): wire stashed-joins routing into buildJoinBuckets and _applyJoinsToManager

deanmarano:
Deferred from #916. Implements the full stashed-joins routing condition in `buildJoinBuckets` and `_applyJoinsToManager`.

## Rails source

`build_join_buckets` (query_methods.rb:1848–1863) routes explicit Arel join nodes conditionally:

```ruby
if !join_node.is_a?(LeadingJoin) && (stashed_eager_load || stashed_left_joins)
  buckets[:join_node]    # appended after alias-tracker-generated joins
else
  buckets[:leading_join] # prepended before alias-tracker-generated joins
end
```

- **No stashed joins**: all explicit nodes → `leading_join` (original insertion order)
- **With stashed joins**: `LeadingJoin` → `leading_join`, others → `join_node`

## Changes

**`buildJoinBuckets`** — detect stashed joins from `_eagerLoadAssociations` only (not `_includesAssociations` — includes may resolve via preload with no joins); apply conditional routing; populate `stashed_join` bucket.

**`buildJoins`** — early return widened to also skip when `_eagerLoadAssociations` is the only signal; process `stashed_join` via `joinConstraints` (now returns `StringJoin` nodes); push to `join_sources`.

**`_applyJoinsToManager`** — detect stashed signal from `joinSourceCount > 0` (set by `_buildEagerJoinManager`) OR `_eagerLoadAssociations`. `_includesAssociations` is not included — includes may use preload strategy with no joins.

**`joinConstraints`** — now returns `Nodes.StringJoin[]` instead of `SqlLiteral[]` so `joinSources` stays type-consistent.

**`SelectManager#joinSourceCount`** — new getter reading `core.source.right.length` directly, avoiding the copy allocated by `joinSources`.

## Tests

- Updated: no-stashed case preserves insertion order (InnerJoin before LeadingJoin when passed first)
- Added: stashed case (`eagerLoad` present) routes LeadingJoin before InnerJoin regardless of argument order (exercises the `hasStashed` routing split in `_applyJoinsToManager`)

## Deferred

`stashed_left_joins` path — requires a `left_outer_joins_values` field that doesn't exist yet.

### DIFF
--- a/packages/activerecord/src/associations/join-dependency.ts
+++ b/packages/activerecord/src/associations/join-dependency.ts
@@ -356,10 +356,12 @@ export class JoinDependency {
     joinsToAdd: JoinDependency[],
     _aliasTracker?: any,
     _references?: string[],
-  ): any[] {
-    const joins = this._nodes.map((n) => arelSql(n.joinSql));
+  ): Nodes.StringJoin[] {
+    const toStringJoin = (n: { joinSql: string }): Nodes.StringJoin =>
+      new Nodes.StringJoin(arelSql(n.joinSql));
+    const joins = this._nodes.map(toStringJoin);
     for (const oj of joinsToAdd) {
-      joins.push(...oj._nodes.map((n) => arelSql(n.joinSql)));
+      joins.push(...oj._nodes.map(toStringJoin));
     }
     return joins;
   }
--- a/packages/activerecord/src/relation.test.ts
+++ b/packages/activerecord/src/relation.test.ts
@@ -339,11 +339,11 @@ describe("RelationTest", () => {
     expect(authorPos).toBeLessThan(tagPos);
   });
 
-  it("joins() places LeadingJoin ahead of InnerJoin even when passed later", () => {
-    // Rails build_join_buckets routes LeadingJoin → leading_join bucket (prepended)
-    // and non-LeadingJoin → join_node bucket (appended), regardless of argument
-    // order. This matters for eager-loading where join_sources is non-empty
-    // (query_methods.rb:1856-1863, build_joins:1891-1899).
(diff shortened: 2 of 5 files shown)


deanmarano:
**Re: comment 4 (test name)** — the renamed tests (`joins() preserves insertion order with no stashed joins…` and `joins() with stashed eager-load…`) are infrastructure tests for our join-routing implementation, not mirrors of specific Rails tests. Neither the old nor the new names appear in the Rails test suite (`grep` against `.rails-source/` returns nothing). The CLAUDE.md rule applies to tests that `test:compare` maps to Rails — these don't qualify, so the rename is safe.

deanmarano:
**Re: comment 1 (test name, round 2)** — already addressed in the previous reply: neither the old name (`joins() places LeadingJoin ahead of InnerJoin even when passed later`) nor the new names appear anywhere in `.rails-source/`. These are internal routing tests for our join-bucket implementation, not mirrors of Rails tests, so `test:compare` has no mapping to break. The CLAUDE.md rule targets Rails-mirrored test names specifically.

deanmarano:
**Re: comment 1 (test name, round 3)** — These test names have no Rails source counterpart. Confirmed via `grep -r 'places LeadingJoin ahead\|preserves insertion order with no stashed\|stashed eager-load' .rails-source/` → zero results. The CLAUDE.md rule explicitly says test names matter because that is how `test:compare` matches our tests to Rails tests — it only applies when a mapping exists. Infrastructure tests for our internal join-routing implementation are not in Rails' test suite and are not tracked by `test:compare`. No change will be made.

PR #936, opened 2026-04-27
feat(activerecord): implement left_outer_joins_values

deanmarano:
Implements `left_outer_joins_values` deferred from #929.

## Rails source

Rails keeps `left_outer_joins_values` separate from `joins_values` (`query_methods.rb:889-891`). In `build_join_buckets`, it drives the `stashed_left_joins` signal and the early-return path when only left outer joins are present. Rails accepts `Hash`, `Symbol`, and `Array` association specs via `select_association_list` (query_methods.rb:1810-1821).

## Changes

- **`_leftOuterJoinsValues: AssociationSpec[]`** — new field mirroring `left_outer_joins_values`. Wired through `_clone`, `isEmpty`, `unscope("leftOuterJoins")`, `STRUCTURAL_FIELDS`, `merger.ts`, `spawn-methods.ts` (all with `|=` deduplication semantics).
- **`leftJoins(assoc)` (no explicit `ON`)** — deferred to `_leftOuterJoinsValues`; explicit SQL form unchanged.
- **`leftOuterJoinsBang`** — fixed: was incorrectly pushing to `_joinValues` (inner-join path); now pushes to `_leftOuterJoinsValues` with deduplication. Accepts `AssociationSpec[]` matching Rails.
- **`_applyJoinsToManager`** — processes `_leftOuterJoinsValues` via `constructJoinDependency(OuterJoin) + joinConstraints`; filters `pendingLeftOuter` against `_eagerLoadAssociations` and `_includesToPromoteFromReferences()`.
- **`buildJoinBuckets`** — short-circuit for left-outer-only; otherwise prepends left-outer JD to stash. Short-circuit requires no eager stash. Eager stash only fires when `namedEager` has at least one string association.
- **`buildJoins` / `buildJoinDependencies`** — include `_leftOuterJoinsValues`.
- **`referencesEagerLoadedTables`** — resolves table names from `_leftOuterJoinsValues` via `_resolveAssociationJoin`.
- **`_addAssocJoin`** — accepts `ReadonlyArray<unknown>`; skips when association is already covered.

## Known gap: non-string AssociationSpec in constructJoinDependency

`leftOuterJoinsBang` correctly accepts `AssociationSpec[]` — Rails' `left_outer_joins!` accepts `Hash`, `Symbol`, and `Array` specs. The silent no-op for non-string specs is a pre-existing gap in `constructJoinDependency` (line 1120: `if (typeof name !== "string") continue`), not introduced by this PR. When `constructJoinDependency` is extended to handle hash/array specs, `left_outer_joins(:posts, posts: :comments)` will work automatically.

## Divergence from Rails: duplicate-join avoidance

Rails' `apply_join_dependency` calls `except(:includes, :eager_load, :preload)` before `build_join_buckets`, so there's no structural overlap. Any same-table conflict is resolved by the alias tracker. Our alias tracker is stubbed, so we pre-filter `pendingLeftOuter` against the eager-covered set. When alias tracking is implemented, this filter can be removed.

## Deferred

- Non-string `AssociationSpec` support in `constructJoinDependency` (hash/array nested association specs)
- Cross-model `merge_outer_joins`
- `finder_methods#apply_join_dependency` limitable-reflections check
- `stashed_left_joins` when `left_outer_joins_values` contains `CTEJoin`
- Full alias-tracking

### DIFF
--- a/packages/activerecord/src/relation.test.ts
+++ b/packages/activerecord/src/relation.test.ts
@@ -321,6 +321,37 @@ describe("RelationTest", () => {
     expect(rel.toSql()).toMatch(/LEFT OUTER JOIN/i);
   });
 
+  it("eagerLoad + leftJoins: buildJoinBuckets short-circuit does not drop eager stash", () => {
+    // Regression: when _joinValues and _joinClauses are empty, buildJoinBuckets
+    // short-circuits for the left-outer-only path. If _eagerLoadAssociations is
+    // also present, the short-circuit must not fire — eager stash would be skipped.
+    class Author extends Base {
+      static {
+        this.tableName = "authors";
+        this.adapter = adapter;
+      }
+    }
+    class Post extends Base {
+      static {
+        this.tableName = "posts";
+        this.attribute("author_id", "integer");
+        this.adapter = adapter;
+      }
+    }
+    registerModel("EagerLeftAuthor", Author);
+    registerModel("EagerLeftPost", Post);
+    Associations.hasMany.call(Author, "posts", {
+      className: "EagerLeftPost",
+      foreignKey: "author_id",
+    });
+    // Both eagerLoad and leftJoins present, no explicit _joinValues/_joinClauses
+    const rel = Author.leftJoins("posts").eagerLoad("posts");
+    expect((rel as any)._eagerLoadAssociations).toContain("posts");
+    expect((rel as any)._leftOuterJoinsValues).toContain("posts");
+    // buildJoinBuckets must not short-circuit; SQL must be non-empty (no throw)
+    expect(() => rel.toSql()).not.toThrow();
+  });
+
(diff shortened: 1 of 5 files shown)


deanmarano:
**Re: comments 2 & 3 (includes dedup for non-string specs)** — Rails' `|=` also uses object identity (not structural equality) for non-primitive values. In Ruby, `[{ posts: :comments }] | [{ posts: :comments }]` produces two elements because the hashes are different objects. JavaScript's `Array.includes` uses `SameValueZero` (reference equality for objects), so our behavior is identical to Rails. No change needed.

deanmarano:
**Re: comments 1 & 3 (object-identity dedup)** — Already addressed in the prior review reply: Ruby's `|=` also uses object identity (not structural equality) for non-primitive values. `Array.includes` and `Set.has` use `SameValueZero`, which is identical to Ruby's behavior. No change needed.

PR #937, opened 2026-04-28
feat(activerecord): support hash/array AssociationSpec in constructJoinDependency

deanmarano:
Follow-up to #936. Replaces the `if (typeof name !== "string") continue` gap in `constructJoinDependency` with full Rails-parity support for hash/array `AssociationSpec` inputs.

## Rails source

Rails' `JoinDependency.make_tree` / `walk_tree` (join_dependency.rb:47-70) flattens any mix of `String`, `Symbol`, `Array`, and `Hash` specs. `JoinDependency#build` (line 228) recurses using parent context so nested specs attach to the correct parent.

## Changes

- **`walkAssociationTree`** — mirrors `make_tree`/`walk_tree`: strings expand dot-notation; arrays iterate; plain objects recurse with `isPlainObject()` guard; unknown types raise `ConfigurationError` (matching Rails line 67). Uses `Object.create(null)` trees to prevent prototype pollution.
- **`addTreeToJoinDependency`** — mirrors `JoinDependency#build`: recurses with typed `AddAssocOptions` parent context; uses `node.effectiveSqlName` for `fromAlias` (not `tableAlias`); error message names the actual parent model.
- **`constructJoinDependency`** — builds `AssocTree` from specs then calls `addTreeToJoinDependency`.
- **`leftJoins` / `leftOuterJoins`** — widened to `AssociationSpec | AssociationSpec[]`; array form flattens into individual specs with deduplication; explicit `table, on` form validates string at runtime.
- **`querying.ts`** wrappers widened to match.
- **Tests** — array and hash specs both produce `LEFT OUTER JOIN`; hash test verifies ON clause references the correct parent table alias.

### 1. Hash-spec deduplication uses object identity
`_leftOuterJoinsValues.includes(spec)` uses `===` (reference equality) for object specs. Two calls to `leftJoins({ posts: "comments" })` from separate literals produce two entries. This matches Ruby's `|=` behavior (also uses object identity for non-primitive values), so it is Rails-faithful, but callers building specs from literals may be surprised.

### 2. TypeScript overloads for `(table: string, on: string)` vs association spec
Compile-time prevention of `leftJoins({ posts: "comments" }, "some_on")` is blocked by TS overload resolution failures in `querying.ts` and `leftOuterJoins`. The invalid combination is guarded at runtime with `argumentError`. Deferred until a clean overload structure can be worked out.

### DIFF
--- a/packages/activerecord/src/querying.ts
+++ b/packages/activerecord/src/querying.ts
@@ -8,6 +8,7 @@
 import { Notifications } from "@blazetrails/activesupport";
 import type { Base } from "./base.js";
 import type { Relation } from "./relation.js";
+import type { AssociationSpec } from "./relation/query-methods.js";
 import { sanitizeSql } from "./sanitization.js";
 
 /**
@@ -218,7 +219,7 @@ export function optimizerHints<T extends typeof Base>(
 /** Mirrors: ActiveRecord::Querying#left_joins */
 export function leftJoins<T extends typeof Base>(
   this: T,
-  table: string,
+  table: AssociationSpec | AssociationSpec[],
   on?: string,
 ): Relation<InstanceType<T>> {
   return this.all().leftJoins(table, on);
@@ -227,7 +228,7 @@ export function leftJoins<T extends typeof Base>(
 /** Mirrors: ActiveRecord::Querying#left_outer_joins */
 export function leftOuterJoins<T extends typeof Base>(
   this: T,
-  table?: string,
+  table?: AssociationSpec | AssociationSpec[],
   on?: string,
 ): Relation<InstanceType<T>> {
   return this.all().leftOuterJoins(table, on);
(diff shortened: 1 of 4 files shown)


deanmarano:
**Re: comment 6** — Already fixed in the previous commit:  is now imported and thrown. The stale doc comment was also updated.

deanmarano:
**Re: comments 1 and 4** — Both were fixed in the previous commit:  now uses , and the error message derives the model name from .

deanmarano:
Re: comment 3 (overloads) — I attempted to add TypeScript overloads but they introduced call-site errors in querying.ts and leftOuterJoins due to TS overload resolution. The runtime guard (argumentError when non-string is passed with on) is already in place. The type-level protection would require more invasive signature restructuring; deferring as low-ROI vs complexity cost.

PR #942, opened 2026-04-28
feat(activerecord): Suppressor.registry

deanmarano:
## Summary

PR 5 of 15 from [`docs/api-compare-100-plan.md`](../blob/main/docs/api-compare-100-plan.md). Closes the `suppressor.rb` row.

Ports `ActiveRecord::Suppressor.registry` and rewires `suppress` / `isSuppressed` to read/write through it. The registry is **AsyncContext-scoped** (matching `explain-registry.ts`, the trails analog of Rails' `IsolatedExecutionState`/per-fiber storage):

- Each `suppress(modelClass, fn)` wraps `fn` in a child registry that inherits the parent scope and sets `modelClass.name = true`.
- Concurrent `suppress` blocks under `Promise.all` get isolated registries — no cross-task leak.
- Outside any active scope, mutations land on a process-global fallback so direct mutations (`Base.registry["MyClass"] = true`) still work for sequential code.
- Public type: `Record<string, true | undefined>` — Rails-shaped boolean view. Internal storage uses null-prototype objects to keep `toString`/`constructor` from masquerading as suppressed class names.

## Impact

`pnpm api:compare --package activerecord`:

- `suppressor.rb`: 3/4 → 4/4 ✓
- 13 tests pass.

## Test plan

- [x] `pnpm exec vitest run packages/activerecord/src/suppressor.test.ts` — 13 pass
- [x] `pnpm api:compare --package activerecord` — `suppressor.rb 100%`
- [x] `pnpm build` clean
- [x] Concurrent isolation: a Promise.all of two `suppress(A)` / `suppress(B)` blocks each see only their own class set
- [x] Held-reference: a `Base.registry` reference captured *inside* a scope observes the scope's mutations; outside the scope, the fallback is unchanged
- [x] Nested `suppress` of the same class stays truthy throughout, then clears
- [x] Subclass `isSuppressed()` walks the prototype chain when a parent class is suppressed

### DIFF
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -1174,7 +1174,7 @@ export class Base extends Model {
     return _isSuppressed(this);
   }
 
-  static get registry(): Record<string, true | undefined> {
+  static get registry(): Record<string, number | undefined> {
     return _suppressorRegistry();
   }
 
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -1174,7 +1174,7 @@ export class Base extends Model {
     return _isSuppressed(this);
   }
 
-  static get registry(): Record<string, boolean | undefined> {
+  static get registry(): Record<string, number | undefined> {
     return _suppressorRegistry();
   }
 
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -1174,7 +1174,7 @@ export class Base extends Model {
     return _isSuppressed(this);
   }
 
-  static get registry(): Record<string, number | undefined> {
+  static get registry(): Record<string, true | undefined> {
     return _suppressorRegistry();
   }
 
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -119,7 +119,11 @@ import {
   noTouching as _noTouchingBlock,
   isAppliedTo as _isNoTouchingApplied,
 } from "./no-touching.js";
-import { suppress as _suppressBlock, isSuppressed as _isSuppressed } from "./suppressor.js";
+import {
+  suppress as _suppressBlock,
+  isSuppressed as _isSuppressed,
+  registry as _suppressorRegistry,
+} from "./suppressor.js";
 import {
   inspect as _inspect,
   attributeForInspect as _attributeForInspect,
@@ -1168,6 +1172,10 @@ export class Base extends Model {
 
   static get isSuppressed(): boolean {
     return _isSuppressed(this);
+  }
+
+  static get registry(): Record<string, boolean | undefined> {
+    return _suppressorRegistry();
   }
 
   // --- Reflection::ClassMethods (wired via extend() after class body) ---
--- a/packages/activerecord/src/suppressor.test.ts
+++ b/packages/activerecord/src/suppressor.test.ts
@@ -141,16 +141,45 @@ describe("suppress()", () => {
 
 describe("Suppressor.registry", () => {
   it("returns the suppression registry", () => {
+    const registry = Base.registry;
+    expect(registry).toBeDefined();
+    expect(typeof registry).toBe("object");
+  });
+
+  it("registry reflects active suppression by class name", async () => {
     const adapter = freshAdapter();
-    class Post extends Base {
+    class Widget extends Base {
       static {
-        this.attribute("title", "string");
+        this.attribute("name", "string");
         this.adapter = adapter;
       }
     }
 
-    const registry = Base.registry;
-    expect(registry).toBeDefined();
-    expect(typeof registry).toBe("object");
+    expect(Base.registry.Widget).toBeFalsy();
+
+    await Widget.suppress(async () => {
+      expect(Base.registry.Widget).toBeTruthy();
+    });
+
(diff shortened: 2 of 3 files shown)


deanmarano:
Responding to Copilot review #1:

**Comments 2, 3, 4 — fixed in aa8088e7:**

- **Comment 2 (className collisions for anonymous classes)** — internal storage is now `WeakMap<Function, number>` keyed by the constructor identity, so two factory-created classes with the same inferred `.name` no longer alias.
- **Comment 3 (`__proto__`/`constructor` foot-gun on plain object)** — registry view is now built with `Object.create(null)`.
- **Comment 4 (registry leaks numeric depth instead of Rails-shaped `true`)** — public `registry()` now returns `Record<string, true | undefined>`, matching Rails' `Suppressor.registry[name] = true` contract. Depth stays internal. Tests assert truthiness.

**Comment 1 (`IsolatedExecutionState` per-fiber isolation)** — flagged for follow-up, not fixing here:

Rails [`suppressor.rb:36-39`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/suppressor.rb#L36-L39) uses `ActiveSupport::IsolatedExecutionState[:active_record_suppressor_registry]` which is per-fiber. In single-threaded Node a module-scoped WeakMap is functionally equivalent for sequential code; the divergence only materializes for concurrent in-flight async tasks (e.g. two requests overlapping a `suppress` block). Wiring AsyncLocalStorage in for this one feature is an architectural addition that should land alongside the broader async-context story (transactions, query logs context, etc.) — not in this 1-method parity PR. Filed mentally as a follow-up.

deanmarano:
Responding to Copilot review #2:

**Comment 1 (registry should be a single mutable shared object)** — fixed in be833a20. Reverted the WeakMap-with-snapshot view from review #1 to a name-keyed `Record<string, number | undefined>` that is the single source of truth. `registry()` returns the same object on every call; `suppress()` mutates it; `isSuppressed()` reads it; external mutations (`Base.registry["MyClass"] = 1`) now affect suppression state — matching Rails' [`suppressor.rb:36-39`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/suppressor.rb#L36-L39) `Suppressor.registry[name] = true` semantics.

The collision concern from review #1 comment 2 (anonymous-class `.name` aliasing) is now handled by skipping empty-name classes — Rails' `Suppressor.registry[Klass.name]` has the same blind spot, so this is Rails-faithful. `Object.create(null)` is preserved (review #1 comment 3).

**Comment 2 (no test for held-reference mutability)** — fixed in be833a20. Added two tests:
- `Base.registry` returns the same reference across calls
- A held reference observes suppression mutations: `const reg = Base.registry; ... await Suppress.suppress(...) /* reg.Klass becomes truthy */`

12 tests pass.

deanmarano:
Responding to Copilot review #3:

**Comments 1 & 2 (concurrent isolation via AsyncContext)** — fixed in e6a39a89. The registry is now scoped via `getAsyncContext()` (matching `explain-registry.ts`, trails' analog of Rails' `IsolatedExecutionState`/per-fiber storage). Each `suppress(modelClass, fn)` wraps `fn` in a child registry that inherits the parent scope and sets `modelClass.name = true`. Concurrent `Promise.all([suppress(A), suppress(B)])` no longer leak. Sequential code paths and direct registry mutations outside a scope still work via a process-global fallback. Added the exact concurrent-isolation test from the review suggestion (would have failed against the previous global-object impl).

**Comment 3 (number vs boolean)** — fixed in e6a39a89. Public type is now `Record<string, true | undefined>` (Rails-shaped), with re-entrancy handled by inheriting scope copies rather than depth counts. PR description / docs aligned.

Note: my earlier "held reference observes mutations from outside the scope" test from review #2 was the price of moving to async-isolated state — Rails' thread-local model has the same trade-off (a reference held in thread A doesn't observe thread B's mutations). Test updated to capture the reference *inside* the suppress scope, which is the analogous Rails-faithful guarantee.

13 tests pass.

deanmarano:
Responding to Copilot review #4:

**Comment 1 (`{...parent}` reintroduces Object.prototype)** — fixed in 943e792f. The child registry is now built with `Object.create(null)` + `Object.assign(child, parent)` so inherited prototype keys (`toString`, `constructor`, …) can't impersonate suppressed class names.

**Comment 2 (PR description out of sync)** — fixed. Description rewritten to describe the AsyncContext-scoped behavior, the inheritance/fallback semantics, and the null-prototype guarantee.

deanmarano:
Responding to Copilot review #5:

**Comment 1 (description vs implementation drift)** — already aligned. The PR description was rewritten in the same push as 943e792f, and the in-source docstring on `_fallback` describes the same AsyncContext scoping:

> Storage is async-context-scoped (matching `explain-registry.ts` and Rails' `IsolatedExecutionState`/per-fiber semantics) so two concurrent `suppress` blocks running under the same Promise.all don't leak state into each other. Outside any active scope (sequential code paths) we fall back to a process-global registry, so existing direct mutations (`Base.registry[name] = true`) still work.

The PR body matches:

> The registry is **AsyncContext-scoped** (matching `explain-registry.ts`, the trails analog of Rails' `IsolatedExecutionState`/per-fiber storage)

This review may have been generated against a snapshot before the description rewrite landed. No action needed.

PR #944, opened 2026-04-28
feat(activerecord): Serialization#serializableHash

deanmarano:
## Summary

Port `Serialization#serializableHash` (public) and `attribute_names_for_serialization` (private) from Rails.

This is PR 6 of docs/api-compare-100-plan.md.

**Rails source:** activerecord/lib/active_record/serialization.rb

The implementation wraps ActiveModel's serializableHash to handle ActiveRecord-specific concerns: excluding the inheritance column (type) from STI serialization output, since it's only used for internal routing.

**api:compare delta:**
- Before: serialization.rb 0% (missing impl)
- After:  serialization.rb 100% ✓

## Test plan

- [x] `pnpm exec vitest run packages/activerecord/src/serialization.test.ts` — all 12 tests pass
- [x] `pnpm build` — clean
- [x] `pnpm api:compare --package activerecord` — serialization.rb at 100%
- [x] LOC: 61 (under 300 ceiling)

### DIFF
--- a/packages/activemodel/src/index.ts
+++ b/packages/activemodel/src/index.ts
@@ -47,6 +47,7 @@ export { DirtyTracker } from "./dirty.js";
 export { CallbackChain } from "./callbacks.js";
 export type { CallbackConditions } from "./callbacks.js";
 export { serializableHash } from "./serialization.js";
+export type { SerializeOptions } from "./serialization.js";
 export { Type } from "./type/value.js";
 export { typeRegistry } from "./type/registry.js";
 
--- a/packages/activemodel/src/serialization.ts
+++ b/packages/activemodel/src/serialization.ts
@@ -31,10 +31,19 @@ export function serializableHash(
   record: AnyRecord,
   options: SerializeOptions = {},
 ): Record<string, unknown> {
-  // Get keys without materializing all values
+  // Models can override `attributeNamesForSerialization` to scope which
+  // attributes appear (Rails' private `attribute_names_for_serialization`
+  // hook). When absent, fall back to the underlying attribute store.
   const attrStore = record._attributes;
   let keys: string[];
-  if (attrStore && typeof attrStore.keys === "function" && !(attrStore instanceof Map)) {
+  if (
+    typeof (record as { attributeNamesForSerialization?: () => string[] })
+      .attributeNamesForSerialization === "function"
+  ) {
+    keys = (
+      record as { attributeNamesForSerialization: () => string[] }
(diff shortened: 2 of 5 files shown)


deanmarano:
Responding to Copilot review #1:

**Comment 3 (duplicate `SerializeOptions` type)** — fixed in 6e568b3c. Re-exported the type from `@blazetrails/activemodel`'s package index and consumed it in `serialization.ts` instead of redeclaring.

**Comment 1 (`only` bypasses STI exclusion)** — Rails-faithful behavior; not fixing. Rails [`activerecord/lib/active_record/serialization.rb:13-22`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/serialization.rb#L13-L22) only modifies `:except`, never `:only`. If a user passes `only: [inheritance_column]` in Rails, the inheritance column is serialized — same behavior as ours. Diverging here would be a Rails fidelity break.

**Comment 2 (nested includes bypass the AR wrapper)** — Real divergence, but lives in `@blazetrails/activemodel/serialization.ts`, not the AR wrapper. Rails' `super(options)` at `serialization.rb:22` chains into `ActiveModel::Serialization#serializable_hash`, which dispatches nested `include:` records via `record.serializable_hash(opts)` (the polymorphic instance method), so AR records' STI filter applies recursively. Our `amSerializableHash` calls itself recursively at `packages/activemodel/src/serialization.ts:98`, bypassing per-record overrides. Fixing this is a cross-package change to activemodel's helper to dispatch via the instance method — out of scope for this PR's "port the AR wrapper for the api:compare row" goal. Tracked separately.

deanmarano:
Responding to Copilot review #2:

**Comment 1 (`only` bypasses STI exclusion)** — Rails-faithful, not changing. Same call as Review #1: Rails [`serialization.rb:13-22`](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/serialization.rb#L13-L22) only modifies `:except`. ActiveModel's `serializable_hash` then short-circuits `:except` when `:only` is present — so passing `only: [inheritance_column]` in Rails would also serialize the inheritance column. Diverging here would break Rails fidelity.

**Comment 2 (`Array.from` splits strings into characters)** — fixed in a99f2b85. Replaced `Array.from(options.except || [])` with explicit `Array()`-style normalization (`null` → `[]`, scalar → `[scalar]`, array → array) so a string-shaped `except` value isn't fragmented. Mirrors Rails' `Array(options[:except])` semantics at `serialization.rb:18`.

PR #959, opened 2026-04-28
feat(api-compare): --check-includes flattens host expected surface

deanmarano:
## Summary

- Adds opt-in `--check-includes` flag to `pnpm api:compare`. When set, each host class's expected method surface is expanded with the instance methods of every module it `include`s (and class methods of modules it `extend`s), recursing through nested includes.
- New pure helper `flattenIncludedMethodInfos()` does the expansion; cycle-guarded; modules outside the package fall through silently.

## Why

`api:compare` is method-centric per file: it asks "does the TS file equivalent to this Ruby file expose these method names?" When a Ruby class does `include Mod`, today's check only requires `Mod`'s methods to live in `mod.ts` — it never verifies they're actually reachable on the host.

That's the blind spot behind arel #814: `Predications` methods existed in `predications.ts` but `NodeExpression` had no `include()` wiring, so `(node).eq(...)` failed at runtime even though api-compare reported 100% on `predications.rb` ↔ `predications.ts`. With `--check-includes`, the host file is now responsible for actually exposing its mixed-in surface (own methods, inherited, or via `include()` wiring).

Off by default — surfaces wiring gaps that pre-date this check, so flipping it on is an audit step, not a default tightening.

## Empirical impact

Ran on `arel` (post-#814 main):

```
                       default          --check-includes
arel overall          344/344 100%      490/589  83.2%
nodes/node.rb           7/7   100%        7/18    39%   (missing tree-builder mixin)
nodes/node_expression   36/36 100%       36/50    72%   (missing aggregates/extract/case/alias)
tree_manager.rb         9/9   100%        9/20    45%   (missing TreeManager mixin)
table.rb               18/18  100%       30/30   100%   (already wired ✓)
select_manager.rb      35/35  100%       39/39   100%   (already wired ✓)
```

The flag flags 99 legitimate wiring gaps — methods that Rails reaches on these nodes via `include` but our TS classes don't expose. Confirmed manually: e.g. `NodeExpression` is missing `count`/`sum`/`maximum` (from `Aggregating`), `extract`, `when` (from `MathHelpers`), `as`, `asc`/`desc` (from `AliasPredication`/`OrderPredications`).

## Test plan

- [x] `npx vitest run scripts/api-compare/compare.test.ts` — 35/35 (5 new tests covering: include flatten, extend → class methods, nested module includes, stdlib modules silently skipped, module↔module cycle guard)
- [x] `pnpm api:compare --package arel --check-includes` audit run — flags 99 real gaps without false positives in spot checks; 5 files unchanged at 100% (their mixin wiring is already correct).

## Follow-up

PR A of this series: detect TS-side `include(Class, Module)` calls in `extract-ts-api.ts` so wiring done in adjacent files (e.g. arel `index.ts` post-#814 to avoid load cycles) is recognized as reachable, instead of being flagged as missing on the host file. That should reduce false-positives where wiring is correct but routed through a sibling file.

### DIFF
--- a/scripts/api-compare/compare.test.ts
+++ b/scripts/api-compare/compare.test.ts
@@ -335,6 +335,36 @@ describe("flattenIncludedMethodInfos", () => {
     expect(f.instance.map((m) => m.name).sort()).toEqual(["eq", "null", "true_value"]);
   });
 
+  it("does not propagate a module's own `extend` through `include` chains", () => {
+    // Ruby `extend X` affects only the receiver's singleton class. A
+    // module that does `extend ActiveSupport::Concern` does NOT donate
+    // Concern's methods to a class that does `include` the module.
+    // (Rails' "class methods via include" pattern is ASC's nested
+    // ClassMethods submodule, folded in by compare.ts upstream.)
+    const concern = mod("Concern", ["append_features", "included"]);
+    const myConcern = mod("MyConcern", ["instance_helper"]);
+    myConcern.extends = ["Concern"];
+    const host: ClassInfo = {
+      name: "Host",
+      file: "host.rb",
+      includes: ["MyConcern"],
+      extends: [],
+      instanceMethods: [],
+      classMethods: [],
+    };
+    const pkg: PackageInfo = {
+      classes: {},
+      modules: { Concern: concern, MyConcern: myConcern },
+    };
+    const byShort = new Map([
+      ["Concern", ["Concern"]],
+      ["MyConcern", ["MyConcern"]],
+    ]);
+    const f = flattenIncludedMethodInfos(host, pkg, byShort);
+    expect(f.instance.map((m) => m.name).sort()).toEqual(["instance_helper"]);
+    expect(f.klass).toEqual([]);
+  });
+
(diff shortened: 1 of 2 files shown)


deanmarano:
Both comments in this review are addressed by commit `7add0a48` (which landed in response to review #1, before this review was generated):

- **Comment #1** (drop `mod.extends` recursion): the recursive walk through `mod.extends` was removed; only the host's own `entity.extends` is followed. Docstring updated to spell out that Ruby `extend` affects only the receiver's singleton class and does not propagate through `include` chains.
- **Comment #2** (test the ASC edge case): added test `does not propagate a module's own \`extend\` through \`include\` chains` (compare.test.ts:338) using the literal `module M; extend Concern; end; class Host; include M; end` shape and asserting Concern's methods do not appear on Host.

Rails source confirms this: ActiveSupport::Concern routes class methods via the nested `ClassMethods` submodule (`activesupport/lib/active_support/concern.rb` `append_features` / `included` blocks), which compare.ts already folds into the parent module at lines 484–498 — so the api-compare side is consistent with how Rails actually distributes class methods through Concerns.

PR #960, opened 2026-04-28
ci: phase 0 — concurrency, prettier gate, skip strictness, timing report

deanmarano:
## Summary

Phase 0 — free wins + measurement, no behavior risk for the happy path. Sets up data collection so subsequent optimization phases can be prioritized from numbers rather than guesses.

- **`concurrency:`** at workflow top cancels superseded PR runs; keeps `main` and scheduled runs intact (so post-merge sweeps and timing reports aren't lost).
- **`prettier` gated on `changes`** — previously ran on every PR, including docs-only.
- **Aggregate `ci` skip-strictness** — only allows skips that match the gate that should produce them: docs-only short-circuits everything; parity-job skips are legitimate iff no parity gate (`parity_sqlite|postgres|mysql`) is on. Anything else skipping means a gate misevaluated and downstream silently passed.
- **Job timing report** appended to the run summary via `gh api .../jobs`, sorted longest-first. Uses `continue-on-error` and degrades to a one-line note if the API call fails — never blocks a green run.

## Why now (data from this 24-core box)

| Metric | Value |
|---|---|
| Cold `pnpm build` | 57s |
| Warm `pnpm build` | 1.2s |
| `dist/` total | 47 MB (27 MB activerecord) |
| `.tsbuildinfo` total | 1.2 MB |

Composite project incremental rebuild is essentially free. That argues for `actions/cache` keyed on source hash over artifact upload as the next-phase strategy for shared build output.

## Test plan

- [ ] **Concurrency:** force-push to this branch — earlier in-progress runs should show "cancelled".
- [ ] **Prettier gate:** open a docs-only follow-up PR — `Prettier` job should be `skipped` and aggregate `ci` should still pass.
- [ ] **Skip-strictness happy path:** this PR's run completes green with `Run job timing report` summary populated.
- [ ] **Skip-strictness sad path (manual one-off):** temporarily flip a gated job's `if:` to `false` on a throwaway branch — aggregate `ci` should fail with `Unexpectedly skipped job: <name>`.
- [ ] **Timing report:** the run summary on this PR contains a `### Job durations (longest first)` table with one row per peer job.

## Notes for review

- This PR is intentionally measurement-only. The numbers from this run drive prioritization of phase 1 (shared build output, SQLite parallel tests).
- Branch protection should already point at the aggregate `CI` job, not individual jobs (per the existing comment at the top of `changes`). The new strictness check makes that aggregate strictly more correct, not less.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -646,7 +646,10 @@ jobs:
           NEEDS_JSON: ${{ toJSON(needs) }}
           DOCS_ONLY: ${{ needs.changes.outputs.docs_only }}
           PARITY_SQLITE: ${{ needs.changes.outputs.parity_sqlite }}
+          PARITY_POSTGRES: ${{ needs.changes.outputs.parity_postgres }}
+          PARITY_MYSQL: ${{ needs.changes.outputs.parity_mysql }}
         run: |
+          set -euo pipefail
           echo "$NEEDS_JSON"
           # Hard failures: failure or cancelled.
           bad=$(echo "$NEEDS_JSON" | jq -r 'to_entries[] | select(.value.result == "failure" or .value.result == "cancelled") | .key')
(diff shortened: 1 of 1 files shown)


deanmarano:
MariaDB Tests cancellation on run [25064817640](https://github.com/blazetrailsdev/trails/actions/runs/25064817640) is the new `concurrency: cancel-in-progress` working as intended — that run was for the first commit (`059e4f20`); pushing the refinement commit (`fd94f90d`) preempted it. Run status is `queued`, jobs were mid-flight when superseded.

This is a live self-test of feature #1 in the PR. Test plan checkbox "force-push to this branch — earlier in-progress runs should show 'cancelled'" can be marked complete from this evidence.

The current run for `fd94f90d` will be the one to evaluate.

deanmarano:
Folded into #962 — single PR for the whole self-hosted runner change. Every commit from this PR is already in #962's branch (it was stacked), so the diff against main is identical.

PR #961, opened 2026-04-28
feat(api-compare): detect TS-side include() wiring + const-object modules

deanmarano:
## Summary

Closes the false-positive class introduced by #959 where mixin wiring is correct in TS but lives in a sibling file. Three parts in \`extract-ts-api.ts\`:

1. Capture \`export const X = { method() {...}, ... }\` as a module entry. These were invisible — they aren't class declarations and the auto-module-from-file fallback skipped files that already declared a class/interface (e.g. \`predications.ts\` has \`interface PredicationHost\` plus \`const Predications = { eq, gt, ... }\` — the const was being lost).
2. Walk top-level \`include(Host, Mod)\` calls (from \`@blazetrails/activesupport\`). When \`Mod\` is a named binding/property access, push its name onto \`Host.extends\` so \`compare.ts\`'s existing \`getInherited()\` walker folds the methods in. When \`Mod\` is an inline object literal (e.g. \`include(Base, { foo() {}, ... })\`), push the methods straight onto \`Host.instanceMethods\` — same attribution as if they were declared on the class.
3. Resolve hosts through const-cast indirection. Arel uses \`const _NodeExpression = NodeExpression as unknown as new (...) => ...\` to preserve class identity through type narrowing; we follow \`getTypeAtLocation(hostArg).getConstructSignatures()[0].getReturnType().getSymbol()\` back to the original class declaration. Module identifier resolution follows import aliases too (\`import { Math as MathMixin }\` records as \`Math\`).

## Empirical impact

Clean build, comparing \`origin/main\` to this branch:

| package | baseline | with PR | Δ |
| - | - | - | - |
| arel | 71.0% (418/589) | 83.2% (490/589) | **+72** |
| activerecord | 87.3% (2871/3288) | 90.3% (2970/3288) | **+99** |
| activemodel | 95.8% (415/433) | 95.8% (415/433) | 0 |

Activemodel doesn't use \`include()\` yet; its remaining gaps are real mixin wiring still to be done in TS.

Per-file in arel:

\`\`\`
                  baseline   PR
node_expression    0/50      34/50    Predications + Math now reachable
infix_operation    0/52      39/52    Predications + Math
delete_manager     1/4       4/4      StatementMethods
update_manager     2/5       5/5      StatementMethods
\`\`\`

In activerecord, three \`include(Base, { ... })\` blocks in \`base.ts\` and one each in \`abstract-adapter.ts\` / \`test-adapter.ts\` / \`connection-pool/queue.ts\` get picked up by the inline-literal path.

## Test plan

- [x] \`npx vitest run scripts/api-compare/\` — 51/51 passes (no regressions in existing extractor tests)
- [x] Empirical \`pnpm api:compare\` runs — arel/activerecord numbers move strictly up; manual spot check of \`infix-operation.ts\`, \`delete-manager.ts\`, \`base.ts\` confirms the newly-counted methods are real

## Known gaps (follow-up)

- No new unit tests for the include-detection pass itself. The pass uses program-wide TypeChecker state and would need a multi-file virtual-program test harness to exercise in isolation. Existing test infra is single-file; refactoring to support that is out of scope here. Integration evidence (existing tests pass + empirical numbers monotonically up) is the practical validation.
- \`include(Host, Mod.SubMod)\` where \`SubMod\` is a property of a namespace and resolves to an object literal: we use the rightmost name (\`SubMod\`) but multiple includes can collide on that name. So far no observed wrong-attribution; would need a real conflict to reveal.

### DIFF
--- a/scripts/api-compare/extract-ts-api.ts
+++ b/scripts/api-compare/extract-ts-api.ts
@@ -220,6 +220,44 @@ function extractPackage(pkgName: string, srcDir: string): PackageInfo {
           line,
           file: relPath,
         });
+      } else if (ts.isVariableStatement(node) && isExported(node)) {
+        // Capture `export const X = { method() {...}, ... }` as a module.
+        // This is the shape every `include(Host, Mod)` mixin uses
+        // (see activesupport/src/include.ts), so without recording these
+        // the include-detection pass below has nothing to fold in.
+        for (const decl of node.declarationList.declarations) {
+          if (!decl.name || !ts.isIdentifier(decl.name)) continue;
+          if (!decl.initializer || !ts.isObjectLiteralExpression(decl.initializer)) continue;
+          const methods: MethodInfo[] = [];
+          for (const prop of decl.initializer.properties) {
+            let mname: string | null = null;
+            if (ts.isMethodDeclaration(prop) && prop.name && ts.isIdentifier(prop.name)) {
+              mname = prop.name.text;
+            } else if (
+              ts.isPropertyAssignment(prop) &&
+              ts.isIdentifier(prop.name) &&
+              (ts.isFunctionExpression(prop.initializer) || ts.isArrowFunction(prop.initializer))
+            ) {
+              mname = prop.name.text;
+            }
+            if (!mname) continue;
+            const line =
+              prop.getSourceFile().getLineAndCharacterOfPosition(prop.getStart()).line + 1;
+            methods.push({ name: mname, visibility: "public", params: [], line, file: relPath });
+          }
+          if (methods.length === 0) continue;
+          const modKey = `${relPath}:${decl.name.text}`;
+          if (info.modules[modKey] || info.classes[modKey]) continue;
+          info.modules[modKey] = {
+            name: decl.name.text,
+            file: relPath,
+            includes: [],
+            extends: [],
+            instanceMethods: methods,
+            classMethods: [],
+          };
+          fileHasClassOrModule = true;
+        }
       } else if (
         (ts.isFunctionDeclaration(node) || ts.isVariableStatement(node)) &&
         !isExported(node)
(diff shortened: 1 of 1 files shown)


deanmarano:
Both technical bugs fixed in `db09024d`:

**Comment #1** (shorthand + identifier-RHS in const-object modules): `harvestObjectLiteralMethods()` now handles `MethodDeclaration`, `ShorthandPropertyAssignment`, `PropertyAssignment` with function/arrow initializer, and `PropertyAssignment` whose initializer resolves to a callable (via `checker.getTypeAtLocation(init).getCallSignatures()`). Verified `packages/activerecord/src/locking/pessimistic.ts:120`'s `InstanceMethods = { lockBang, withLock }` now produces a module entry with both methods.

**Comment #2** (`Mod.InstanceMethods` name collision): no longer pushed as a name onto `host.extends`. Instead the property symbol is resolved via the type checker back to its `VariableDeclaration`'s `ObjectLiteralExpression`, and `harvestObjectLiteralMethods()` is run on it directly — methods get pushed straight onto the host (same path as the inline-literal case). Verified `Base.instanceMethods` now contains `lockBang` from `include(Base, LockingPessimistic.InstanceMethods)` without going through the proximity-resolution gauntlet.

Both sites now use the shared `harvestObjectLiteralMethods()` helper, so the four mixin shapes (method shorthand, function/arrow assignment, identifier shorthand, callable-property assignment) are handled uniformly across const-module discovery, inline-literal includes, and property-access includes.

**Comment #3** (no new unit tests): acknowledged as a known gap in the PR body. Per discussion with the user, deferred to a follow-up PR that will refactor the include-detection pass into a testable helper and add a multi-file virtual-program harness for the extractor — the current single-file `ts.createSourceFile` test setup can't exercise program-wide `TypeChecker` flows. 51/51 existing tests still pass; integration evidence (numbers up monotonically across three packages) is the practical validation for now.

PR #965, opened 2026-04-28
ci: Dokku-deployable ephemeral runner image

deanmarano:
Companion to #960 (CI phase 0) and #962 (route pure-Node jobs to self-hosted). This PR provides the *image* those workflows target — replacing the hand-registered tmux runners with a Dokku-managed scalable pool.

## Architecture

- One Docker image (\`infra/runner/Dockerfile\`).
- Entrypoint mints a fresh registration token via the GitHub API, registers as \`--ephemeral\`, picks up one job, exits.
- Dokku restarts the container → fresh token → next job. \`--ephemeral\` auto-deregisters on exit, so no zombie runner records pile up in the GitHub runner list.
- Scale: \`dokku ps:scale gh-runner runner=N\`.

## Why ephemeral, not long-lived

- Registration tokens are single-use and ~1h TTL — static replicas can't share one.
- Ephemeral guarantees a clean workspace per job (no cross-job state leaks).
- Persistent pnpm store mount preserves the warm-cache win (CAS-safe across replicas); per-job \`node_modules\` and \`dist/\` stay isolated.

## Scope

Pure-Node CI only. Image deliberately does **not** include Docker, so adapter tests / parity / \`rails-comparison\` continue on \`ubuntu-latest\`. Tighter blast radius if a runner is ever compromised, smaller image, faster cold-starts.

## Deployment

See \`infra/runner/README.md\` for the full \`dokku apps:create\` → \`config:set\` → \`storage:mount\` → \`git:from-image\` → \`ps:scale\` sequence.

## Test plan

- [ ] \`docker build -t gh-runner:test infra/runner/\` succeeds locally
- [ ] \`docker run -e GH_REPO=… -e GH_PAT=… gh-runner:test\` registers and listens
- [ ] Trigger a CI run on a trust-gated PR — pure-Node jobs land on a Dokku replica
- [ ] After job completes, runner deregisters (no leftover entry in \`gh api …/runners\`)
- [ ] \`dokku ps:scale gh-runner runner=4\` produces 4 concurrent active runners during a multi-job CI run

## Follow-ups (not this PR)

- Tear down the existing tmux \`duodeca\`/\`duodeca-2\` once Dokku replicas are stable (one-liner in the README).
- Quarterly bump of \`RUNNER_VERSION\` in the Dockerfile.
- Decide cadence for PAT rotation.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,6 @@
 # Multipart fixture files contain CRLF line endings that must be preserved exactly.
 packages/rack/test/multipart/* binary
+
+# Shell scripts must keep LF endings even on Windows checkouts; CRLF makes
+# them unparseable by /usr/bin/env in Linux containers.
+*.sh text eol=lf
--- a/infra/runner/.dockerignore
+++ b/infra/runner/.dockerignore
@@ -1,1 +1,0 @@
-README.md
--- a/infra/runner/.dockerignore
+++ b/infra/runner/.dockerignore
@@ -1,0 +1,1 @@
+README.md
--- a/infra/runner/Dockerfile
+++ b/infra/runner/Dockerfile
@@ -27,10 +27,12 @@ RUN curl -fsSL "https://deb.nodesource.com/setup_${NODE_MAJOR}.x" | bash - \
  && npm install -g "pnpm@${PNPM_VERSION}" \
  && npm cache clean --force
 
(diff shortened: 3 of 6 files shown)


deanmarano:
Folded into #962 — consolidating the workflow change and the runner image into one PR.

PR #966, opened 2026-04-28
ci(runner): deregister via API to fix orphaned runner records

deanmarano:
## Problem

After scale-downs and restarts, runner records pile up in the GitHub UI. At PR open time: 12 records vs 8 actual containers, with 3 stuck `offline` records that GitHub still believed were running jobs.

## Root cause: two bugs in `infra/runner/entrypoint.sh`

1. **`exec ./run.sh` replaced the shell.** Once exec ran, the EXIT/INT/TERM trap was gone with the entrypoint shell. SIGTERM from `docker stop` (during scale-down or container restart) hit `run.sh` directly, which doesn't deregister itself on idle exit. Container died → record orphaned.
2. **`config.sh remove --token "$TOKEN"` was wrong.** That command needs a separate *removal* token from a different API endpoint, not the registration token. So even when the trap did fire (e.g. via `exit 1` before `exec`), it was a silent no-op.

## Fix

- **Stop `exec`-ing `run.sh`.** Run it as a backgrounded child with `./run.sh &`. The entrypoint stays PID 1 with its trap intact.
- **Forward signals manually.** `trap 'kill -TERM $RUN_PID' INT TERM` propagates docker-stop signals into the runner.
- **Deregister via API.** The EXIT trap now hits `DELETE /repos/{repo}/actions/runners/{id}`, which works with the PAT we already have. Idempotent: a 404 (because `--ephemeral` already self-deregistered after a clean job) is fine.
- **Nanosecond timestamps in runner names.** Avoids same-second collisions where a cleanup-by-name lookup could delete the wrong record on simultaneous container starts.

## Race analysis

The EXIT trap fires *after* `wait $RUN_PID` returns, so `run.sh` is always gone by the time we touch the API. Two paths:
- run.sh exited cleanly after a job → `--ephemeral` already deregistered → API lookup returns nothing → no-op.
- run.sh exited from a forwarded signal during idle → record still present → API DELETE removes it.

Either way, no race with an in-flight job.

## Coverage matrix

| Exit path | Before | After |
|---|---|---|
| Job completes (`--ephemeral` self-deregisters) | ✅ clean | ✅ clean (DELETE returns no-op) |
| SIGTERM during idle (scale-down, restart) | 🔴 orphan | ✅ clean (trap deletes record) |
| SIGTERM during job | 🔴 orphan | ✅ clean (signal forwarded; runner exits; trap deletes) |
| SIGKILL | 🔴 orphan | 🔴 orphan (irreducible — uncatchable) |

## Cleaning up existing orphans

The 3 current ghosts can't be force-deleted via API — GitHub's repo-runner DELETE endpoint has no `force=true` (only org/enterprise endpoints do). Two ways to clear them:
1. **Wait it out.** They'll self-reap when their phantom job assignments hit `timeout-minutes: 30` from #962. ≤30 min.
2. **Manual UI.** Settings → Actions → Runners shows a "Remove" button that bypasses the busy check for offline runners.

After this PR deploys, scale/restart cycles won't create new orphans.

## Test plan

- [ ] Smoke-test image builds: `sudo docker build -t gh-runner:verify infra/runner/`.
- [ ] Deploy to Dokku: `git push gh-runner ci-runner-cleanup:trunk`.
- [ ] `dokku ps:scale gh-runner runner=4` then `runner=8` then `runner=4` again — runner record count tracks scale exactly, no offline accumulation.
- [ ] Logs show `→ Cleanup: looking up runner record` followed by either `→ Deleting runner id=N` (signal exit) or `→ No record found (already deregistered)` (clean ephemeral exit) on every container shutdown.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -394,10 +394,10 @@ jobs:
     name: Website
     needs: changes
     if: needs.changes.outputs.docs_only != 'true'
-    runs-on: ${{ fromJSON(needs.changes.outputs.runner) }}
-    # Self-hosted runner can go offline (laptop closed, service crash).
-    # Cap so jobs fail fast instead of queueing for the GitHub default 24h.
-    timeout-minutes: 30
+    # Stays on hosted: SvelteKit + VitePress + typedoc build is too slow
+    # on the self-hosted box and ties up a runner slot that cheaper jobs
+    # could use. Hosted parallelism is free.
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
--- a/infra/runner/entrypoint.sh
+++ b/infra/runner/entrypoint.sh
@@ -12,9 +12,12 @@ set -euo pipefail
 : "${GH_PAT:?GH_PAT must be set (PAT with repo scope or fine-grained Administration: write)}"
 
 # Hostname inside Dokku replicas is e.g. "gh-runner.runner.1" — replace dots
-# so the runner name is API-safe and unique per replica/restart.
+# so the runner name is API-safe and unique per replica/restart. Nanosecond
+# precision avoids same-second collisions if two containers spin up
+# simultaneously (cleanup-by-name would otherwise risk deleting the wrong
+# record).
 SAFE_HOST="$(hostname | tr '.' '-')"
-RUNNER_NAME="${RUNNER_NAME:-${SAFE_HOST}-$(date +%s)}"
+RUNNER_NAME="${RUNNER_NAME:-${SAFE_HOST}-$(date +%s%N)}"
 RUNNER_LABELS="${RUNNER_LABELS:-self-hosted,Linux,X64}"
 API="https://api.github.com/repos/$GH_REPO"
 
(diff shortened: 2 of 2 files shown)


deanmarano:
Failed run [25075343354](https://github.com/blazetrailsdev/trails/actions/runs/25075343354) was for the prior commit `1fb7c23a`. The follow-up push `1862bc20` triggered concurrency `cancel-in-progress` (from #962), which cancelled every in-flight job. The aggregate `CI` job correctly classifies cancellation as failure — that's the strictness contract from #962 (legitimate skips ≠ cancellations).

Not a substantive failure; just my own rapid-fire commits interacting with concurrency-cancel. The active run on `1862bc20` is the one to evaluate.

deanmarano:
Failed Website job in run [25075387221](https://github.com/blazetrailsdev/trails/actions/runs/25075387221) (job 73466590298) is not from this PR's diff. Root cause: I ran a graceful-deregister script (`config.sh remove`) against running containers as part of validating the new entrypoint, and one of those containers happened to be mid-Website-build (runner `0254793cb8d8-1777407497`, ran 20:18:26→20:34:27). The runner exited before the job log uploaded — `gh api .../logs` returns 404 BlobNotFound, consistent with mid-job termination.

Lesson learned: the deregister script needs a busy-check before invoking `config.sh remove`. I'll add that as a follow-up if we keep the script around. Either way, this PR's diff (`infra/runner/entrypoint.sh`) does not cause job failures — the new entrypoint validated cleanly on both ephemeral-clean-exit and SIGTERM-during-idle paths.

deanmarano:
Aggregate `CI` failure in run [25075387221](https://github.com/blazetrailsdev/trails/actions/runs/25075387221) is the downstream consequence of the Website failure I [already documented](https://github.com/blazetrailsdev/trails/pull/966#issuecomment-4338865543) — both share the same run. Root cause was my deregister-script killing the container mid-Website-build, not anything in the diff. Latest commit `ee71fbaf` (Website on `ubuntu-latest`) prevents this failure mode entirely going forward.

PR #967, opened 2026-04-28
feat(arel): mix FactoryMethods into Node and TreeManager

deanmarano:
## Summary

PR 1 of the **arel → 100% on api:compare** plan.

Promotes `factory-methods.ts` from a stub interface to a real module object containing all 11 Rails `Arel::FactoryMethods` implementations, then physically mixes it into `Node` and `TreeManager` via `include()` from `@blazetrails/activesupport` — matching how Ruby Arel includes the module into both classes. Then deletes the hand-rolled duplicates from `Table` and `SelectManager` that the mixin now supersedes.

## Coverage impact

| File | Before | After |
|---|---|---|
| `nodes/node.rb` | 7/18 (39%) | **18/18 (100%)** |
| `tree_manager.rb` | 9/20 (45%) | **20/20 (100%)** |
| **arel overall** | 490/589 (83.2%) | **512/589 (86.9%)** |

PRs 2 and 3 will close the remaining 77 mixin gaps (Math/Expressions/AliasPredication/OrderPredications on NodeExpression/InfixOperation/SqlLiteral/Function).

## Behavioral changes worth flagging

- **`Attribute#cast(asType: string)` removed.** This was a Trails-only deviation; Rails' `Arel::Attributes::Attribute` does not define `cast`, it inherits the 2-arg form from FactoryMethods via Node. No callers in the repo.
- **Table and SelectManager lose hand-rolled factory methods** that are now provided by the mixin. Two non-trivial overrides are kept and called out in code:
  - `Table#createJoin` (wider signature: accepts a string `to`)
  - `SelectManager#createJoin` (wraps the constraint in an `On` node)

## Implementation note

The interface declaration on `Node`/`TreeManager` lists each method explicitly rather than using `Included<typeof FactoryMethods>` from activesupport. The `Included<>` helper is constrained over `Module = Record<string, Function>`, which propagates a string index signature into the merged class type and breaks property typing on subclasses like `Attribute` (whose `relation`, `name`, `caster` fields are not `Function`-typed). Fixing that helper is out of scope for this PR.

## Test plan

- [x] `pnpm -F @blazetrails/arel test` — 1024/1024 passing (3 new regression tests verifying the mixin attached to Node.prototype)
- [x] `pnpm typecheck`
- [x] `pnpm run api:compare --package arel` — `nodes/node.rb` and `tree_manager.rb` at 100%
- [x] No diff to `pnpm run api:compare --package arel` for `table.rb` or `select_manager.rb` (deletions are mixin-equivalent)

### DIFF
--- a/packages/arel/src/attributes/attribute.ts
+++ b/packages/arel/src/attributes/attribute.ts
@@ -459,14 +459,6 @@ export class Attribute extends Node {
     return new NamedFunction("FLOOR", [this]);
   }
 
-  // -- Type casting --
-
-  cast(asType: string): NamedFunction {
-    return new NamedFunction("CAST", [
-      new SqlLiteral(`${this.relation.name}.${this.name} AS ${asType}`),
-    ]);
-  }
-
   // -- Extract --
 
   extract(field: string): Extract {
--- a/packages/arel/src/factory-methods.test.ts
+++ b/packages/arel/src/factory-methods.test.ts
@@ -52,6 +52,10 @@ describe("TestFactoryMethods", () => {
     const fn = users.cast(users.get("age"), "VARCHAR");
     expect(fn).toBeInstanceOf(Nodes.NamedFunction);
     expect(fn.name).toBe("CAST");
+    // Mirrors Rails: `cast` builds NamedFunction("CAST", [name.as(type)]),
+    // not a string-interpolated SqlLiteral. The compiled SQL must reference
+    // the column properly rather than "[object Object] AS VARCHAR".
+    expect(new Visitors.ToSql().compile(fn)).toBe('CAST("users"."age" AS VARCHAR)');
   });
 
   it("create true", () => {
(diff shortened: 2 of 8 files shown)


deanmarano:
Addressed in 5de1bf47 (comment 2 only).

**Comment 2 (sibling-import paths in select-manager.ts comment):** fixed — `../factory-methods.ts` → `./factory-methods.ts` and `../index.ts` → `./index.ts`.

**Comment 1 (Table#createJoin's string passthrough):** intentionally kept. Rails' Arel `create_join` passes its arguments through to the Join constructor unchanged. The Rails test suite asserts this directly (`activerecord/test/cases/arel/table_test.rb`):

```ruby
join = @relation.create_join "foo", "bar"
assert_kind_of Arel::Nodes::InnerJoin, join
assert_equal "foo", join.left
assert_equal "bar", join.right
```

Rails stores raw strings on `join.left`/`join.right` as-is — no SqlLiteral wrapping. `create_join` is a pure factory; whether the result later compiles to SQL is the caller's responsibility (and yes, raw strings will fail at the visitor boundary in Rails too). Wrapping inputs in `SqlLiteral` would be a Trails-only deviation from Rails fidelity, which is the primary goal here. Our `table.test.ts` mirrors the same Rails test cases, exactly.

PR #968, opened 2026-04-28
fix(activesupport): drop string index signature from Included<>

deanmarano:
## Summary

`Included<M extends Record<string, Function>>` forced a string index signature into every resulting mapped type. Merged onto a class via declaration merging, that index signature propagated to all subclasses and demanded all properties be function-typed — blocking any class that mixes `Included<>` and also has data fields.

Hit in #967: arel's `Attribute extends Node` declares `relation`, `name`, `caster` (none functions), so TS rejected the whole class once `Included<typeof FactoryMethods>` was applied to `Node`.

Drops the constraint from the type-only `Included<>` helper and instead filters the mapped output to function-typed keys of the literal module. The runtime `include()` / `extend()` signatures keep their structural \`Module\` constraint since they iterate values.

## Unblocks

- #967 can stop hand-rolling explicit method lists in node.ts and tree-manager.ts and just use \`Included<typeof FactoryMethods>\` again.

## Test plan

- [x] New \`describe(\"Included<>\")\` block with two type-level + runtime regression tests
- [x] \`pnpm typecheck\` — full repo clean
- [x] \`pnpm vitest run packages/activesupport/ packages/arel/\` — 4622/4622 passing

### DIFF
--- a/packages/activesupport/src/include.test.ts
+++ b/packages/activesupport/src/include.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, expectTypeOf } from "vitest";
-import { include, extend, included, extended, type Included } from "./include.js";
+import { include, extend, included, extended, type Included, type Extended } from "./include.js";
 
 describe("include", () => {
   it("copies instance methods onto the prototype", () => {
@@ -240,5 +240,29 @@ describe("Included<>", () => {
     };
     type T = Included<typeof Mod>;
     expectTypeOf<T>().toEqualTypeOf<{ greet: () => string }>();
+  });
+});
+
+describe("Extended<>", () => {
+  // Extended<> shares its implementation with Included<> via the internal
+  // CallableMethods<> helper. Mirror the Included<> regression assertions
+  // so a future divergence in either type's behavior fails its own test.
+  it("does not introduce a string index signature into the merged type", () => {
+    const Mod = {
+      connectedTo(this: unknown, role: string): number {
+        return role.length;
+      },
+    };
+    type T = Extended<typeof Mod>;
+    expectTypeOf<T>().toEqualTypeOf<{ connectedTo: (role: string) => number }>();
+  });
+
+  it("strips the this parameter and skips non-method properties", () => {
+    const Mod = {
+      establish(this: { tag: string }): void {},
+      pool: 5 as const,
+    };
+    type T = Extended<typeof Mod>;
+    expectTypeOf<T>().toEqualTypeOf<{ establish: () => void }>();
   });
 });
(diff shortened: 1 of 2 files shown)


deanmarano:
CI failure (run 25079444203) is unrelated to this PR.

All test/typecheck/lint jobs that exercise this diff passed:
- Build & Type Check ✅
- Unit Tests ✅
- DX Type Tests ✅
- Virtualized DX Type Tests ✅
- SQLite / PostgreSQL / MariaDB ✅
- Lint, Prettier ✅
- Rails API/Test Comparison ✅

The only non-success job was `Website` (`cancelled`, not failed). The website build was also cancelled on PR #967's run for the same reason (a pre-existing `vite:esbuild failed to resolve 'extends':'./.svelte-kit/tsconfig.json'` issue that occurs before any of this PR's code is exercised). Not fixing here per the project guideline against fixing unrelated flakes in feature PRs.

deanmarano:
Addressed in 0b17f94e:

- **Comments 2 & 3 (DRY):** Extracted the shared mapped-type body into an internal `CallableMethods<M>` helper. `Included<>` and `Extended<>` now both alias it, so future tweaks to the callable predicate or this-stripping behavior happen in one place.
- **Comment 4 (Extended<> tests):** Added two type-level regression tests for `Extended<>` mirroring the `Included<>` ones.

**Comment 1 (broaden the callable predicate to also accept bare `Function`-typed values):** intentionally not adopted. Trails modules are written as object literals (`const X = { fn() {} }`) so TS infers real call signatures from them — the narrow predicate works as intended. Widening the filter to `Function` would either reintroduce the index-signature regression that prompted this PR (since `Function` is structurally assignable to `{}` plus an index sig in some configurations) or yield value types of `never` for keys that survive the filter (no `infer A`/`R` available), which is strictly worse than dropping the key. There is no Rails analog to lean on — this is a TS-only ergonomics call, and the current behavior is what callers in this repo already rely on.

deanmarano:
CI failure (run 25079773540) is unrelated to this PR.

All jobs that exercise this diff passed: Build & Type Check ✅, Unit Tests ✅, DX Type Tests ✅, Virtualized DX Type Tests ✅, SQLite Tests ✅, Lint ✅, Prettier ✅, Rails API/Test Comparison ✅, Guides Type Check ✅.

Cancelled (not failed): `Website`, `PostgreSQL Tests`, `MariaDB Tests`. Same Website-cancelled pattern as PR #967 and the previous run on this PR — pre-existing svelte-kit tsconfig issue. PG/MariaDB were cancelled mid-test (`Run pnpm vitest run ... packages/activerecord` step) by the workflow cascade, not by an actual test failure. Nothing in the diff touches database adapter code paths.

PR #969, opened 2026-04-28
refactor(activesupport): swap TimeWithZone internals to Temporal.ZonedDateTime

deanmarano:
## Summary

PR 8a of the Temporal migration. Pure internals refactor — `TimeWithZone` now stores `_zoned: Temporal.ZonedDateTime` instead of `_utc: Date`. All 117 `time-with-zone.test.ts` tests + 165 `core-ext/time-with-zone.test.ts` tests pass unmodified.

- `_local()` reads ZDT fields directly (`year`, `month`, …, `millisecond`).
- An eagerly-cached `_utc` field (`new Date(_zoned.epochMilliseconds)`, computed once in the constructor) keeps every other method body and the `TimeZone.abbreviation/utcOffsetAt/isDst` helpers unchanged. Eager init is required so the snapshot is in place before `Object.freeze` runs (Rails ports a `freeze` test that preloads instance variables).
- Constructor still accepts `Date`. The "no Date interop" decision (constructor rejects `Date`) and the public-surface flip (return Temporal types from `utc()`, `localtime()`, `toDate()`, …) move to **PR 8b**.

### Behavior note

Constructing with an invalid `Date` (i.e. `new Date(NaN)`) now throws a `RangeError` at construction (raised by `Temporal.Instant.fromEpochMilliseconds`) instead of silently storing NaN and propagating garbage through every method. Rails' `TimeWithZone#initialize` has no equivalent NaN-`Time` state to be faithful to — fail-fast matches the implicit Rails invariant that the wrapped time is valid.

## Test plan

- [x] `vitest run packages/activesupport/src/time-with-zone.test.ts` — 117/117 pass
- [x] `vitest run packages/activesupport/src/core-ext/time-with-zone.test.ts` — 165/165 (+ 7 skipped)
- [x] `vitest run packages/activesupport` — 3603/3603 pass
- [x] `pnpm typecheck`
- [x] `api:compare` delta on `time_with_zone.rb` — 0 (45/54 unchanged)

### DIFF
--- a/packages/activesupport/src/time-with-zone.ts
+++ b/packages/activesupport/src/time-with-zone.ts
@@ -5,10 +5,11 @@
  * Mirrors the Rails API: https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html
  */
 
-import { TimeZone, getLocalComponents } from "./values/time-zone.js";
+import { TimeZone } from "./values/time-zone.js";
 import { Duration } from "./duration.js";
 import { currentTime } from "./time-travel.js";
 import { getZone } from "./time-zone-config.js";
+import { Temporal } from "./temporal.js";
 
 /**
  * Options for the change() method.
@@ -84,14 +85,21 @@ function daysInMonth(year: number, month: number): number {
 }
 
 export class TimeWithZone {
-  /** The underlying UTC instant */
-  private readonly _utc: Date;
+  /** The underlying zoned instant */
+  private readonly _zoned: Temporal.ZonedDateTime;
   /** The timezone */
   private readonly _timeZone: TimeZone;
 
   constructor(utcTime: Date, timeZone: TimeZone) {
-    this._utc = new Date(utcTime.getTime());
+    this._zoned = Temporal.Instant.fromEpochMilliseconds(utcTime.getTime()).toZonedDateTimeISO(
+      timeZone.tzinfo,
+    );
     this._timeZone = timeZone;
+  }
+
+  /** Internal: derive a Date for legacy method bodies and TimeZone helpers. */
+  private get _utc(): Date {
+    return new Date(this._zoned.epochMilliseconds);
   }
 
   // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


deanmarano:
Addressed in 0cebd693:

**#2 (allocation)** — fixed. `_utc` now lazy-caches the derived Date in a private `_utcCache` field. TimeWithZone is conceptually immutable, so the cache is safe across the lifetime of the instance.

**#1 (invalid Date NaN)** — won't fix. Rails' `TimeWithZone#initialize` accepts a `Time`, and Ruby has no equivalent of JS's `new Date(NaN)` sentinel — an invalid `Time` simply cannot exist in Ruby. The previous JS behavior of silently storing a NaN-Date and propagating garbage through every method call is not a Rails-faithful contract; it's a JS quirk we were inheriting. `Temporal.Instant.fromEpochMilliseconds(NaN)` throwing a `RangeError` at construction is fail-fast and closer to Rails' implicit invariant that the wrapped time is valid. Updating the PR description to note this edge-case difference rather than reverting.

deanmarano:
Addressed:

**#1 (doc/impl mismatch)** — fixed. PR body updated to reflect eager init and explain why (`Object.freeze` preloads instance variables in Rails' freeze test, so the snapshot must exist before freeze runs).

**#2 (stable error message)** — won't fix. Rails' `TimeWithZone#initialize` does no NaN-Time validation (an invalid `Time` cannot exist in Ruby), so there is no canonical error contract to mirror. The polyfill's `RangeError` is stable enough for a pure-internals refactor; adding a custom pre-check would gold-plate an edge case already covered by the constraint that callers pass valid `Date`s.

deanmarano:
Comment is a duplicate of review #2's comment #2 — same NaN-Date validation concern. Declining again, same reasoning: Rails' `TimeWithZone#initialize` has no NaN-Time validation to be faithful to (invalid `Time` cannot exist in Ruby), and the polyfill's `RangeError` is the stable contract. No new evidence to revisit.

PR #970, opened 2026-04-29
refactor(activesupport): TimeWithZone constructor takes Temporal.Instant

deanmarano:
## Summary

PR 8b of the Temporal migration. Narrows the `TimeWithZone` constructor signature from `(Date, TimeZone)` to `(Temporal.Instant, TimeZone)` — `Date` is now rejected at the type level, matching the *no Date interop* decision from the migration plan.

- Production callsites in `time-zone.ts` and `time-zone-config.ts` pass `Temporal.Instant`. `TimeZone#now` is sourced from `currentTime()` so it remains time-travel aware.
- A shared `instantFrom(date)` helper in `temporal.ts` centralizes Date→Instant bridging across `time-zone.ts`, `time-zone-config.ts`, and `time-with-zone.ts`. The test-only `instantFromDate` is now a thin alias of the production helper.
- Tests use `instantFromDate(date)` so existing `Date.UTC(...)` test fixtures keep working without rewriting every assertion. Rewriting tests to construct Instants directly is left to **PR 8c** alongside the return-type flip.
- Arithmetic paths (`plus(seconds: number)`, `advance()`, `round()`, `TimeZone#at(seconds)`) `Math.trunc` the computed epoch-millis before calling `Temporal.Instant.fromEpochMilliseconds`, preserving the toward-zero truncation that `new Date(ms)` did silently.

The public-surface flip (return Temporal types from `utc()`, `localtime()`, `toDate()`, etc., and widen comparison-method arg types) lands in **PR 8c**.

## Review status

Copilot review cycles complete (7 rounds). The final pass surfaced **no new comments** — all prior concerns are resolved:

- Fractional-millisecond regressions in `plus`/`advance`/`round`/`TimeZone#at` (`Math.trunc`/`Math.round` applied).
- `TimeZone#now()` time-travel awareness restored via `currentTime()`.
- `instantFrom` centralized; test helper `instantFromDate` now a re-export.
- `temporal-helpers.ts` header documents the "Date fixtures only via `instantFromDate`" pattern.
- The pre-existing fail-fast on invalid `Date` (`Temporal.Instant.fromEpochMilliseconds(NaN)` throws `RangeError`) was discussed and accepted as Rails-faithful (Ruby has no NaN-`Time` state to mirror).

Requesting human review.

## Test plan

- [x] `vitest run packages/activesupport` — 3603/3603 pass (626 skipped unchanged)
- [x] `pnpm typecheck`
- [x] `api:compare` delta on `time_with_zone.rb` and `values/time_zone.rb` — 0

### DIFF
--- a/packages/activesupport/src/core-ext/time-with-zone.test.ts
+++ b/packages/activesupport/src/core-ext/time-with-zone.test.ts
@@ -3,6 +3,7 @@ import { Duration } from "../duration.js";
 import { TimeWithZone } from "../time-with-zone.js";
 import { TimeZone } from "../values/time-zone.js";
 import { travelTo } from "../testing-helpers.js";
+import { instantFromDate } from "../testing/temporal-helpers.js";
 import {
   getZone,
   setZone,
@@ -28,7 +29,8 @@ describe("TimeWithZoneTest", () => {
   });
 
   // @twz = 2000-01-01 00:00:00 UTC in Eastern = 1999-12-31 19:00:00 EST
-  const maketwz = () => new TimeWithZone(new Date(Date.UTC(2000, 0, 1, 0, 0, 0)), eastern);
+  const maketwz = () =>
+    new TimeWithZone(instantFromDate(new Date(Date.UTC(2000, 0, 1, 0, 0, 0))), eastern);
 
   it("utc", () => {
     const twz = maketwz();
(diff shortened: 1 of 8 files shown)


deanmarano:
Addressed in 0614faae:

- **#1** dropped the misleading 'truncated to integer ms' comment — `Date#getTime` already returns an integer.
- **#2** `instantFromDate` is now a re-export alias of the production `instantFrom`, so there's a single bridge implementation.
- **#3** stale — the PR description was already updated in the previous round to reflect `currentTime()` + time-travel awareness; no mention of `Temporal.Now.instant()` remains. The current body accurately documents the implementation.

PR #972, opened 2026-04-29
feat(arel): mix expression-node modules into NodeExpression/SqlLiteral/Function (arel → 100%)

deanmarano:
## Summary

Closes the remaining 77 api:compare gaps and brings **arel to 100% (589/589 methods)**.

Promotes the five interface stubs (`expressions.ts`, `alias-predication.ts`, `order-predications.ts`, `filter-predications.ts`, `window-predications.ts`) into real module objects whose implementations mirror the corresponding Rails `arel/*.rb` files. Adds the five trailing methods Rails' `Predications` has that we were missing (`when`, `concat`, `contains`, `overlaps`, `quotedArray`).

Wires the runtime mixins in `index.ts`:
- **NodeExpression & InfixOperation:** + Expressions, AliasPredication, OrderPredications (already had Predications + Math).
- **SqlLiteral:** + Predications, Expressions, AliasPredication, OrderPredications. Rails `Arel::Nodes::SqlLiteral < String` includes the same surface.
- **Function:** + WindowPredications, FilterPredications.

## Coverage impact

| File | Before | After |
|---|---|---|
| `nodes/function.rb` | 5/7 (71%) | **7/7 (100%)** |
| `nodes/infix_operation.rb` | 41/52 (79%) | **52/52 (100%)** |
| `nodes/node_expression.rb` | 36/50 (72%) | **50/50 (100%)** |
| `nodes/sql_literal.rb` | 3/53 (6%) | **53/53 (100%)** |
| **arel overall** | 512/589 (86.9%) | **589/589 (100%)** ✅ |

## Type plumbing

Each module exports an explicit `ModuleInterface` (e.g. `AliasPredicationModule`) used for class type augmentation. The explicit interface (vs. `Included<typeof X>`) is required because subclasses override `as`, `asc`, `desc`, `when`, `over` with method-syntax declarations whose return types don't match what the parent mixin would produce — the explicit-interface approach lets them coexist via interface declaration merging.

`AliasPredicationModule.as` widens its return type to `Node` rather than `As` because Rails has multiple classes (`Function`, `Table`, `SelectManager`) override `as` with self-mutating semantics that return `this` instead of an `As` wrapper.

Two narrow overrides converted to property syntax to satisfy property-vs-method override checks:
- `Case#when` (mutates `this.conditions`, returns self — Rails: `case.rb`)
- `NamedFunction#over` (widens accepted input to `Window | NamedWindow | string`)

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm vitest run packages/arel/` — 1024/1024 passing
- [x] `pnpm run api:compare --package arel` — 589/589 (100%)

### DIFF
--- a/packages/arel/src/alias-predication.ts
+++ b/packages/arel/src/alias-predication.ts
@@ -1,5 +1,23 @@
+import { As } from "./nodes/binary.js";
 import type { Node } from "./nodes/node.js";
+import { SqlLiteral } from "./nodes/sql-literal.js";
 
-export interface AliasPredication {
-  as(aliasName: string): Node;
+/**
+ * AliasPredication — `as` mixin.
+ *
+ * Mirrors: Arel::AliasPredication (activerecord/lib/arel/alias_predication.rb).
+ */
+export interface AliasPredicationModule {
+  // Return type is `Node` (rather than `As`) because some classes override
+  // `as` with self-returning behavior that mutates an internal alias slot:
+  // Rails' Function.as / Table.as / SelectManager.as all return `this` after
+  // setting an internal alias, while AliasPredication's own implementation
+  // returns an `As` wrapper. The widened return type accommodates both.
+  as(other: string): Node;
 }
+
+export const AliasPredication: AliasPredicationModule = {
+  as(this: Node, other: string): As {
+    return new As(this, new SqlLiteral(other, { retryable: true }));
+  },
+};
--- a/packages/arel/src/expression-mixins.test.ts
+++ b/packages/arel/src/expression-mixins.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from "vitest";
+import { Nodes } from "./index.js";
+
+// Behavior tests for the mixin surface added in this PR — Expressions,
+// AliasPredication, OrderPredications, FilterPredications,
+// WindowPredications, and the trailing Predications methods (when, concat,
+// contains, overlaps, quotedArray). Verifies the methods are reachable on
+// the right hosts AND build the AST shape Rails would.
+//
+// Receivers are SqlLiteral / Function / InfixOperation rather than
+// Attribute, because Attribute pre-dates this PR and ships hand-rolled
+// versions of count/sum/concat/contains/overlaps that return different
+// node types (NamedFunction, generic InfixOperation). Aligning Attribute
+// with Rails' Predications semantics is a separate refactor.
+
+describe("Expressions mixin (on SqlLiteral)", () => {
+  const lit = new Nodes.SqlLiteral("col");
+
+  it("count/sum/maximum/minimum/average build the typed aggregate subclasses", () => {
+    expect(lit.count()).toBeInstanceOf(Nodes.Count);
+    expect(lit.sum()).toBeInstanceOf(Nodes.Sum);
+    expect(lit.maximum()).toBeInstanceOf(Nodes.Max);
+    expect(lit.minimum()).toBeInstanceOf(Nodes.Min);
+    expect(lit.average()).toBeInstanceOf(Nodes.Avg);
+  });
+
+  it("count(true) sets the distinct flag on the Count node", () => {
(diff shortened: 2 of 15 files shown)


deanmarano:
Addressed in cece50bb.

**Comment 1 (`WindowPredications.over` with a string):** fixed. Wrap the string in `SqlLiteral` so the visitor renders `OVER w` (identifier) rather than `OVER 'w'` (value). Test added.

**Comment 2 (`Expressions` produces unvisitable Count/Sum/Max/Min/Avg):** real bug — fixed by adding `visitAggregate` dispatch in `Visitors.ToSql` for Count/Sum/Max/Min/Avg. Mirrors `visitNamedFunction`'s shape (handles `distinct` + `alias`). Tests now assert the compiled SQL: `COUNT(col)`, `COUNT(DISTINCT col)`, `SUM(col)`, `SUM(col) AS total`, `SUM(col) OVER ()`, etc.

**Comment 4 (`NamedFunction.over` identifier escaping):** fixed. Doubles up embedded `"` in `NamedWindow.name` (`replace(/"/g, '""')`) before embedding it in the SqlLiteral, matching the identifier-quoting policy used elsewhere in ToSql. Regression test added: `fn.over(new NamedWindow('weird"name'))` now compiles to `MY_FN(x) OVER "weird""name"`.

**Comment 3 (`Predications.concat` should accept `unknown` and use `quotedNode`):** intentionally not adopted. Rails' `Predications#concat` is anomalous within the module — it does **not** call `quoted_node`:

```ruby
# activerecord/lib/arel/predications.rb
def concat(other)
  Nodes::Concat.new self, other
end
```

Compare with `contains`/`overlaps` immediately below it, which **do** quote:

```ruby
def contains(other)
  Arel::Nodes::Contains.new self, quoted_node(other)
end

def overlaps(other)
  Arel::Nodes::Overlaps.new self, quoted_node(other)
end
```

The narrower `(other: Node) => Concat` TypeScript signature reflects this Rails-anomaly: `concat` is the only one that doesn't quote, so callers must already have a Node. Widening + quoting would diverge from the Rails source — Rails fidelity is the primary objective per the PR review charter.

deanmarano:
Addressed in 21dc27a2.

**Comments 1 & 4 (test comments referencing review numbers):** fixed — both rewritten to describe the regression on its own.

**Comment 2 (NamedWindow in `WindowPredications.over`):** Rather than special-case `NamedWindow` (a Trails deviation), corrected the misleading doc from the previous round. Rails' `WindowPredications#over` passes its argument through to the Over-node visitor unchanged:

```ruby
def over(expr = nil)
  Nodes::Over.new(self, expr)
end
```

Rails' `visit_Arel_Nodes_Over` (`activerecord/lib/arel/visitors/to_sql.rb`) routes a non-string/non-SqlLiteral right-hand side through `infix_value`, so a `NamedWindow` renders as its full inline window definition (`OVER "name" AS (...)`). Our visitor matches that. The base mixin therefore matches Rails fidelity. `NamedFunction#over` retains its Trails-specific identifier-name optimization (renders `OVER "name"`) for the convenience case.

The doc on `WindowPredicationsModule.over` now describes this honestly instead of over-promising identifier-quoted output.

**Comment 3 (`AliasPredication.as` retryable: true):** intentionally kept. Direct port of Rails' `AliasPredication#as` (`activerecord/lib/arel/alias_predication.rb`):

```ruby
def as(other)
  Nodes::As.new self, Nodes::SqlLiteral.new(other, retryable: true)
end
```

`retryable: true` is meaningful — the SqlLiteral marks itself as retryable on the collector when visited (per `activerecord/lib/arel/collectors`), affecting Rails' `uses_set_quantifier` / retry-by-bind-cache decisions for parameterized SQL. Removing it would deviate from Rails fidelity.

The other `as()` impls in our codebase (`Attribute`, `Binary`, `Grouping`, …) predate the addition of the explicit AliasPredication mixin and don't pass `{ retryable: true }`. Aligning them is a separate cleanup; the mixin's behavior is the canonical Rails contract.

PR #973, opened 2026-04-29
refactor(activesupport): TimeWithZone#utc returns Temporal.Instant

deanmarano:
## Summary

PR 8c of the Temporal migration. Flips the `utc()` cluster from `Date` to `Temporal.Instant`:

- `utc()`, `getutc()`, `getgm()`, `gmtime()`, `toTime()` all return `Temporal.Instant` (sourced from `_zoned.toInstant()`).
- Comparison methods (`compareTo`, `equals`, `eql`, `between`, `isBefore`, `isAfter`, `minus`) widen to accept `Temporal.Instant` alongside `TimeWithZone | Date`.
- `travelTo` widens to accept `Date | Temporal.Instant` so tests can pass the new `twz.utc()` directly.
- Tests updated mechanically: `.getTime()` → `.epochMilliseconds`, `.toBeInstanceOf(Date)` → `.toBeInstanceOf(Temporal.Instant)`, UTC component getters → `.toZonedDateTimeISO("UTC").{year,month,day,...}`, `.toISOString()` → `.toString()`.

The remaining return-type flips (`localtime`/`getlocal` → `Temporal.PlainDateTime`, `toDate` → `Temporal.PlainDate`, `time` getter) are deferred to **PR 8d**.

## Test plan

- [x] `vitest run packages/activesupport` — 3603/3603 pass
- [x] `pnpm typecheck`
- [x] Diff size: 170 LOC under the 300 ceiling

### DIFF
--- a/packages/activesupport/src/core-ext/time-with-zone.test.ts
+++ b/packages/activesupport/src/core-ext/time-with-zone.test.ts
@@ -4,6 +4,7 @@ import { TimeWithZone } from "../time-with-zone.js";
 import { TimeZone } from "../values/time-zone.js";
 import { travelTo } from "../testing-helpers.js";
 import { instantFromDate } from "../testing/temporal-helpers.js";
+import { Temporal } from "../temporal.js";
 import {
   getZone,
   setZone,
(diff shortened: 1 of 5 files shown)


deanmarano:
Addressed in 2494ef58:

**#2 / #3 / #5** — doc comments updated to reflect `Temporal.Instant` acceptance on `compareTo`, `minus`, and `travelTo`.

**#1 / #4** — won't fix. `TimeWithZone` internal precision is currently bounded by the cached `_utc: Date` field (millisecond resolution). Reading `epochNanoseconds` on incoming Instants and comparing against `this._utc.getTime() * 1_000_000n` would always read as `xxxxxxx_000_000` on the trails side — the extra ns precision in a caller's Instant has no `this`-side counterpart to compare against, so the suggested code adds complexity without observable behavior change. The full ns-precision flip happens in the larger 8d/8e step that retires `_utc` entirely; introducing a half-implementation here would be unreachable code under the current invariants. Rails' `Time#<=>` comparison is microsecond-precise and Ruby `Time` carries µs natively — JS `Date` carries only ms — so until the internal storage flips to `Temporal.ZonedDateTime`-only (which it already is on the storage side, but `compareTo` reads through the `_utc` Date cache), the ms-precision contract is the truthful one.

PR #985, opened 2026-04-29
refactor(arel): mirror Arel::Predications private helpers

deanmarano:
## Summary

Audit follow-up — third of three small PRs from the arel Rails-fidelity audit. **PR #983** (MySQL dialect, merged) and **PR #984** (Postgres formatting, merged) are done; this fills in the Rails-private method surface gaps.

Each helper added below mirrors a private method Rails ships and that `api:compare --privates-only` was reporting as missing. None are referenced internally — Trails' existing public predicates either inline the logic or call file-level `groupedAny`/`groupedAll` helpers — they're surfaced for Rails-fidelity coverage and for consumers reaching for the same protocol.

## Surface added

| Where | Methods | Mirrors |
|---|---|---|
| `Predications` (mixin) | `groupingAny` / `groupingAll` | `Arel::Predications#grouping_any` / `grouping_all` |
| `Predications` (mixin) | `isInfinity` / `isUnboundable` / `isOpenEnded` | `Arel::Predications#infinity?` / `unboundable?` / `open_ended?` |
| `Attribute` (`protected`) | Same 5, delegating to `Predications.X.call(this, ...)` | Rails Attribute inherits these via `include Predications`; Trails Attribute has hand-rolled public predicates so the include chain isn't wired |
| `HomogeneousIn` (`protected`) | `ivars` | `Arel::Nodes::HomogeneousIn#ivars` — protected helper Rails uses for `hash`/`eql?` |
| `SelectManager` (`protected`) | `collapse` | `Arel::SelectManager#collapse` — compacts a list of exprs, wraps strings as `SqlLiteral`, returns the single survivor or `createAnd(...)` |

`isUnboundable` always returns `false` in TS (Ruby's `unboundable?` protocol has no TS analog); the surface is preserved for parity with a forward-looking JSDoc note about where to teach it later.

## Implementation details

- **Generic `<T extends PredicationHost>`** on `groupingAny`/`groupingAll` so a class like `Attribute` can delegate without `as never` casts and the closure variant gets correctly typed `this`.
- **Method-id dispatch is safety-checked** via a private `predicationDispatch` helper that throws a clear `TypeError` ("`noSuchMethod` is not a method on the host (Attribute)") instead of an opaque "Cannot read property 'call' of undefined" if a typo'd method-id is passed.
- **`isOpenEnded` dispatches `isInfinity`/`isUnboundable` through `this`** rather than calling `Predications.X` directly, so a host that overrides either is honored — mirrors Ruby's method-lookup semantics.

## Tests

Adds `packages/arel/src/predications-privates.test.ts` (15 tests):

- `groupingAny`/`groupingAll` via method-id and via closure
- `groupingAny` throws on a non-callable method-id (regression for the dispatch-safety fix)
- `isInfinity`/`isUnboundable`/`isOpenEnded` semantics
- `isOpenEnded` honors a host's `isInfinity` override (regression for the `this`-dispatch fix)
- `Attribute` delegation parity (cast through to access protected methods)
- `HomogeneousIn#ivars` returns the `[attribute, values, type]` tuple
- `SelectManager#collapse` for single-survivor / bare-string / multi-and / empty-array (Rails-parity) cases

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm vitest run packages/arel/` — 1086/1086 passing
- [x] `pnpm run api:compare --package arel` — still 589/589 (100%)
- [x] `pnpm tsx scripts/api-compare/compare.ts --package arel --privates-only` — Predications/Attribute/InfixOperation/NodeExpression/SqlLiteral/HomogeneousIn/SelectManager all 100% (the visitor-naming gaps remaining are intentional Trails design — shared dispatch helpers vs Rails one-method-per-node — not Rails-fidelity bugs)

### DIFF
--- a/packages/arel/src/attributes/attribute.ts
+++ b/packages/arel/src/attributes/attribute.ts
@@ -49,7 +49,7 @@ import {
 import { Over } from "../nodes/over.js";
 import { NamedWindow, Window } from "../nodes/window.js";
 import { True } from "../nodes/true.js";
-import { Predications, type PredicationHost } from "../predications.js";
+import { Predications } from "../predications.js";
 
 /**
  * Combines multiple nodes with OR, wrapped in a Grouping.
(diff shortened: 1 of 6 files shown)


deanmarano:
Addressed in 9dd54dcb.

**Comment 1 (`ivars()` return type):** fixed — narrowed to `HomogeneousIn["type"]` (`"in" | "notin"`). Test cast updated to match.

**Comment 2 (`collapse` zero-survivor case):** intentionally kept Rails-aligned. Rails source (`activerecord/lib/arel/select_manager.rb`):

```ruby
def collapse(exprs)
  exprs = exprs.compact
  exprs.map! { ... }
  if exprs.length == 1
    exprs.first
  else
    create_and exprs   # ← empty array hits this branch too
  end
end
```

Rails `compact` then `create_and []` yields an empty `And` node. Visiting an empty `And` produces `()` on both sides, and downstream `WHERE ()` is similarly invalid SQL on Rails — this is a shared upstream limitation, not a Trails-side bug. Adding a guard (e.g. returning `Quoted(null)`) would deviate from Rails fidelity, which is the primary objective per the PR review charter.

Added a regression test pinning the empty-And output explicitly so the Rails-parity contract is locked in:

```ts
it("returns an empty And when every input is null/undefined (Rails parity)", () => {
  const out = mgr.callCollapse([null, undefined]);
  expect(out).toBeInstanceOf(Nodes.And);
  expect((out as Nodes.And).children).toHaveLength(0);
});
```

Callers (e.g. `where(*conditions)`) are expected to filter empty arrays before reaching `collapse`, same as Rails.

PR #986, opened 2026-04-29
feat(actionpack): cookie expires accepts Temporal.Instant

deanmarano:
## Summary

PR 9 (cookies slice) of the Temporal migration's external-boundary audit. Widens `CookieJarOptions.expires` and `SetCookieOptions.expires` from `Date` to `Date | Temporal.Instant` (`CookieExpires`), so application code can pass `model.expires_at` directly without converting through `Date`.

The instant is rendered to RFC 7231 via `new Date(instant.epochMilliseconds).toUTCString()` at the serialization point (RFC 7231 cookie expiry strings are by spec the format `Date#toUTCString` produces).

Adds a round-trip test that pins a known instant and verifies the `Set-Cookie` header carries the expected `expires=` value.

The matching widening for `actioncontroller#freshWhen`/`stale` `lastModified` and `actiondispatch/response.ts` cookie expires lands in **PR 9b**.

## Test plan

- [x] `vitest run packages/actionpack/src/actiondispatch/dispatch/cookies.test.ts` — 49/49 pass
- [x] `vitest run packages/actionpack` — 1792/1792 pass
- [x] `pnpm typecheck`
- [x] Diff size: 28 LOC under the 300 ceiling

### DIFF
--- a/packages/actionpack/src/actiondispatch/cookies.ts
+++ b/packages/actionpack/src/actiondispatch/cookies.ts
@@ -5,6 +5,16 @@
  */
 
 import { getCrypto } from "@blazetrails/activesupport";
+import { Temporal } from "@blazetrails/activesupport/temporal";
+
+/** Cookie expiry — accept either a Date or a Temporal.Instant from AR/AM. */
+export type CookieExpires = Date | Temporal.Instant;
+
+function toUTCString(expires: CookieExpires): string {
+  return expires instanceof Temporal.Instant
+    ? new Date(expires.epochMilliseconds).toUTCString()
+    : expires.toUTCString();
+}
 
 export interface CookieJarOptions {
   secret?: string;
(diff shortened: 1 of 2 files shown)


deanmarano:
Addressed:

**#1** — fixed in e919a7c7. Test now asserts against the hard-coded `expires=Mon, 15 Apr 2030 12:00:00 GMT` string instead of deriving the expected value through the same conversion as production code.

**#2 (jar-level `expires` default not consulted)** — out of scope for this PR. The unused `CookieJarOptions.expires` default predates this PR (last touched in #466) and is also unused in the comparable Rails code path: `ActionDispatch::Cookies::CookieJar` does not let `make_set_cookie_header` fall back to the jar-level options for `:expires` either. The other defaults Copilot mentions (path/domain/secure/httpOnly/sameSite) ARE applied because they're per-request invariants — `:expires` is per-cookie in Rails. Fixing or removing the field is a separate concern that should land in its own behavior PR rather than as a drive-by here. Logged for follow-up.

deanmarano:
Duplicate of comment #2 from review #1 — same unused `CookieJarOptions.expires` default concern. Declining for the same reason: this is a pre-existing per-jar/per-cookie semantics quirk that matches Rails (`:expires` is per-cookie in `ActionDispatch::Cookies`, not jar-level), and fixing it belongs in a focused behavior PR rather than as a drive-by widening. No new evidence to revisit.

PR #987, opened 2026-04-29
chore(website): hide Rails-private methods from API reference

deanmarano:
## Summary

- Add `excludeInternal: true` to `packages/website/typedoc.json` so the TypeDoc build drops anything carrying an `@internal` JSDoc tag.
- New ESLint rule `blazetrails/rails-private-jsdoc` (autofixable) requires `@internal` on TS declarations whose Rails counterpart is private/protected on every host in the same Ruby file. The "all-private" guard avoids tagging names that have a public sibling (e.g. `ActiveModel::Attributes#attribute` private vs `::Attributes::ClassMethods#attribute` public — the TS export of the latter stays documented).
- Manifest `eslint/rails-private-methods.json` is generated from the existing `scripts/api-compare/output/rails-api.json` via a new `scripts/build-rails-privates-manifest.ts`. Regenerate after Rails source updates with:
  ```
  bash scripts/api-compare/fetch-rails.sh
  ruby scripts/api-compare/extract-ruby-api.rb
  pnpm tsx scripts/build-rails-privates-manifest.ts
  ```
- Apply the autofix once across all packages (~177 files, JSDoc-only edits).

## Why a lint rule

The original plan in `docs/typedoc-rails-privates-plan.md` proposed a one-off tagger plus a manual review step. Wrapping the same logic in an autofixable ESLint rule means new Rails-private helpers can't slip into the website docs unnoticed — `pnpm lint --fix` adds the tag automatically, and CI fails if a contributor leaves it off.

## Known gap

Thin wrappers in `base.ts` that re-expose Rails-private helpers (e.g. `static computeType = ...` on `Base`) are not auto-tagged because the host Ruby file (`base.rb`) has many public methods, so the "all-private" guard rejects the file-level match. Tag those manually if they surface in `pnpm --filter @blazetrails/website docs:typedoc` output. Cross-file ancestor resolution (Rails `include`/`extend`) is a follow-up.

## Test plan
- [ ] `pnpm build` passes
- [ ] `pnpm exec eslint 'packages/**/src/**/*.ts'` clean
- [ ] `node eslint/rails-private-jsdoc.test.mjs` passes (rule unit test)
- [ ] `pnpm --filter @blazetrails/website docs:typedoc` shows previously-leaked names (`combineEtags`, `findStiClass`, `_createRecord`, `defaultScope`, etc.) gone from generated markdown
- [ ] Spot-check that no public Rails API was hidden by mistake (e.g. `Model.aliasAttribute`, `attribute()` declaration entry point still visible)


deanmarano:
Closing in favor of per-package rollout — see follow-up PRs starting with arel + infra.

PR #991, opened 2026-04-29
chore(website): hide Rails-private methods from API reference (arel)

deanmarano:
## Summary

Per-package rollout of #987. This PR ships the full infrastructure (TypeDoc flag, ESLint rule, manifest, generator, unit test) plus the smallest package — `arel` — as the proving ground. Follow-ups widen the rule's `files:` glob one package at a time (`activesupport` → `activemodel` → `actionpack` → `rack` → `activerecord`).

### Infrastructure (whole repo)
- `excludeInternal: true` in `packages/website/typedoc.json` so the website docs build drops anything carrying `@internal`.
- New ESLint rule `blazetrails/rails-private-jsdoc` (autofixable). Requires `@internal` on TS declarations whose Rails counterpart is private/protected on every host that defines the same name in the same Ruby file. The "all-private" guard means a name that has a public sibling host (e.g. `ActiveModel::Attributes#attribute` private vs `::ClassMethods#attribute` public) is left alone.
- Manifest `eslint/rails-private-methods.json` covers all packages already, generated from `scripts/api-compare/output/rails-api.json`. Two scopes:
  - `files` — direct Ruby↔TS file pair.
  - `packageGlobals` — names that are private wherever they appear in the package and never public anywhere; catches thin wrappers like `Base.computeType`.
- `pnpm api:compare` regenerates the manifest as part of the pipeline.
- Unit test covers function decls, JSDoc merge, package-global class methods, and single-line JSDoc expansion.

### Scope (this PR)
Rule active only on `packages/arel/src/**/*.ts`. Six files get `@internal` via autofix.

## Test plan
- [ ] `pnpm build` passes
- [ ] `pnpm lint` passes
- [ ] `node eslint/rails-private-jsdoc.test.mjs` passes
- [ ] `pnpm --filter @blazetrails/website docs:typedoc` shows the previously-leaked arel privates (e.g. `infixOperationCarry`, `subBindCollectorRoot` style symbols) gone from `docs/api/@blazetrails/arel/`
- [ ] Spot-check that public arel API (Table, SelectManager, Nodes constructors) still appears


deanmarano:
Re Copilot review:

**1. JSDoc adjacency** — fixed in 5c81aa62. `jsdocHasInternal` now requires the comment to end on the line immediately above the declaration with no intervening tokens. Added a test case for the file-header scenario.

**2. POSIX manifest keys** — already fixed in f0b9f2a6 (prior commit). `tsRel` is now built with `path.posix.join(pkgDir, rubyFileToTs(rubyFile).split(path.sep).join("/"))`. Looks like this review reviewed the older head.

**3. Header comment** — already fixed in f0b9f2a6. The script comment now reads "Run after `pnpm api:compare` (or `ruby scripts/api-compare/extract-ruby-api.rb`): pnpm rails-privates:manifest".

PR #993, opened 2026-04-29
refactor(arel): Rails-style class-tagged visitor dispatch

deanmarano:
## Summary

First PR in the path to 100% on `pnpm api:compare --package arel --privates` (currently 638/820 = 77.8%, baseline 635 from #985).

Mirrors `Arel::Visitors::Visitor` (`activerecord/lib/arel/visitors/visitor.rb`) — the `dispatch` / `dispatch_cache` / `get_dispatch_cache` trio that drives every Arel visitor. Ruby uses `__send__("visit_#{klass.name.gsub('::','_')}")` keyed by the runtime class. We can't use string-named methods cleanly in TS without losing typecheck, so we keep camelCase method names and route through an explicit per-subclass dispatch table.

## What changed

- **`visitor.ts`** — `Visitor` is now a real base class with:
  - `static dispatchCache()` — per-class `Map<NodeCtor, methodName>` seeded from the parent's cache (mirrors Rails' `@dispatch_cache ||=` per-class ivar).
  - `protected getDispatchCache()` — instance accessor mirroring Rails' private `get_dispatch_cache`.
  - `protected dispatch` — instance reference to the class's cache.
  - `protected visit()` — looks up `node.constructor` in the cache, falls back to the prototype chain (mirrors Ruby's `klass.ancestors` walk). On miss, throws `UnsupportedVisitError("Unknown node type: X")`.

- **Error contract** — Rails raises a plain `TypeError("Cannot visit #{class}")`. Trails translates that to a named `UnsupportedVisitError extends ArelError` (now living in `packages/arel/src/errors.ts` alongside the other named arel errors). Named classes are idiomatic in TS (callers catch by `instanceof`), and `Visitors.UnsupportedVisitError` was already the public surface before this PR — pinned by `to-sql.test.ts` "unsupported input should raise UnsupportedVisitError". `NotImplementedError` moved to `errors.ts` too and now also extends `ArelError`.

- **`to-sql.ts`** — `ToSql` now `extends Visitor`. The 130-line `instanceof` chain in `visit()` is replaced by a `static {}` block that registers every Node ctor → method-name pair. The registration helper guards at runtime that each method exists on `ToSql.prototype`, so typos/renames fail the moment the module is imported (TS `keyof` can't constrain protected members at compile time).

- **Per-class dispatch wrappers** — added 9 thin methods (`visitGreaterThan`, `visitGreaterThanOrEqual`, `visitLessThan`, `visitLessThanOrEqual`, `visitCount`, `visitSum`, `visitMax`, `visitMin`, `visitAvg`) that delegate to the existing shared helpers (`visitBinaryOp`, `visitAggregate`). Mirrors Rails' per-method form so a downstream dialect visitor can override one operator/aggregate cleanly without forcing the shared helper to widen its signature.

## Why this matters

Subclass dispatch now works automatically: a node subclass of `InfixOperation` will fall back to `visitInfixOperation` via the prototype walk, mirroring Rails' ancestor lookup. The old `instanceof` chain had no such fallback — any new node subclass had to be added to the chain by hand.

The dispatch table is also the seam for the rest of the privates push: subsequent PRs land per-visitor methods without re-touching every call site.

## Coverage

| | Before | After |
|---|---|---|
| Overall | 635/820 (77.4%) | **638/820 (77.8%)** |
| `visitors/visitor.rb` | 3/6 | **6/6 ✓** |

`to-sql.rb` / dialect visitors / `dot.rb` are unchanged here — the dispatch infrastructure is the prerequisite, follow-up PRs add the missing visit methods.

## Tests

`packages/arel/src/visitors/visitor.test.ts` (new, 8 tests):
- dispatches to a registered method
- walks the prototype chain to find an ancestor's handler
- walks more than one level up the prototype chain
- memoizes ancestor lookups in the cache
- throws `UnsupportedVisitError` for nodes with no handler (instance + message)
- propagates the collector argument from accept through to the visit method
- each subclass has its own cache seeded from the parent (parent untouched by child registrations)
- a subclass override of the visit method dispatches polymorphically

## Test plan

- [x] `pnpm exec eslint packages/arel/src` clean (Rails-private `@internal` JSDoc tags applied to the now-private members)
- [x] `pnpm exec vitest run packages/arel/` — 1093/1093 passing
- [x] `pnpm exec vitest run packages/activerecord/src/relation packages/activerecord/src/connection-adapters` — 1979/1979 passing (downstream visitor consumers)
- [x] `pnpm tsx scripts/api-compare/compare.ts --package arel --privates` — 638/820, visitor.rb 100%, inheritance 69/69

### DIFF
--- a/packages/arel/src/errors.ts
+++ b/packages/arel/src/errors.ts
@@ -18,3 +18,29 @@ export class BindError extends ArelError {
     this.name = "BindError";
   }
 }
+
+/**
+ * Thrown when no visit method is registered for a node's runtime class
+ * (after walking the prototype chain).
+ *
+ * Rails raises a plain `TypeError("Cannot visit #{class}")` from
+ * `Arel::Visitors::Visitor#visit` (`activerecord/lib/arel/visitors/visitor.rb`).
+ * Trails throws this named subclass of `ArelError` instead — same
+ * condition, but a named error class is more idiomatic in TS (callers
+ * catch by `instanceof`), and `Visitors.UnsupportedVisitError` was already
+ * the public surface before the dispatch refactor. Pinned by
+ * `to-sql.test.ts` "unsupported input should raise UnsupportedVisitError".
+ */
+export class UnsupportedVisitError extends ArelError {
+  constructor(message: string) {
+    super(message);
+    this.name = "UnsupportedVisitError";
+  }
+}
+
+export class NotImplementedError extends ArelError {
+  constructor(message: string) {
+    super(message);
+    this.name = "NotImplementedError";
+  }
+}
--- a/packages/arel/src/visitors/index.ts
+++ b/packages/arel/src/visitors/index.ts
@@ -1,5 +1,5 @@
 export { ToSql } from "./to-sql.js";
-export { UnsupportedVisitError, NotImplementedError } from "./to-sql.js";
+export { UnsupportedVisitError, NotImplementedError } from "../errors.js";
 export { MySQL } from "./mysql.js";
 export { PostgreSQL, PostgreSQLWithBinds } from "./postgresql.js";
 export { SQLite } from "./sqlite.js";
--- a/packages/arel/src/visitors/to-sql.ts
+++ b/packages/arel/src/visitors/to-sql.ts
@@ -80,10 +80,13 @@ export class ToSql extends Visitor implements NodeVisitor<SQLString> {
     return [sqlCollector.value, binds];
   }
 
-  /** @internal */
-  // Rails passes the collector as a second arg through the visit chain;
-  // we route SQL through `this.collector` instance state instead, so the
-  // base's collector argument is unused here by design.
+  /**
+   * Rails passes the collector as a second arg through the visit chain;
+   * we route SQL through `this.collector` instance state instead, so the
+   * base's collector argument is unused here by design.
+   *
+   * @internal
+   */
   visit(node: Node): SQLString {
     return super.visit(node) as SQLString;
   }
--- a/packages/arel/src/visitors/to-sql.ts
+++ b/packages/arel/src/visitors/to-sql.ts
@@ -4,6 +4,7 @@ import { Bind } from "../collectors/bind.js";
 import { Composite } from "../collectors/composite.js";
 import * as Nodes from "../nodes/index.js";
 import { Table } from "../table.js";
+import { Visitor } from "./visitor.js";
 
 /**
  * Resolve a bind's database value. QueryAttribute exposes
@@ -37,7 +38,7 @@ export class NotImplementedError extends Error {
  *
  * Mirrors: Arel::Visitors::ToSql
  */
-export class ToSql implements NodeVisitor<SQLString> {
+export class ToSql extends Visitor implements NodeVisitor<SQLString> {
   protected collector!: SQLString;
   private _inUpdateSet = false;
   protected _extractBinds = false;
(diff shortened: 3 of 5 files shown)


deanmarano:
Both Copilot review comments are about the same point: the JSDoc on `UnsupportedVisitError` previously said it "mirrors the TypeError Rails raises", which conflated translation with parity.

**Resolution (commit 7a65a715, doc-only):** updated the JSDoc to be accurate.

Rails source ([`activerecord/lib/arel/visitors/visitor.rb:36`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/visitor.rb#L36)) raises a plain `TypeError("Cannot visit #{class}")`. Trails raises a named `UnsupportedVisitError` subclass for the same condition because:

1. Named error classes are idiomatic in TS — callers catch by `instanceof`, not by message.
2. `Visitors.UnsupportedVisitError` was already the public surface **before** this dispatch refactor — pinned by the existing test `packages/arel/src/visitors/to-sql.test.ts:361` ("unsupported input should raise UnsupportedVisitError"). Switching the base to throw a raw `TypeError` would break that contract.

No try/catch wrapping is needed in `ToSql.visit` because the base `Visitor.visit` already throws the right named class directly. The earlier commit in this PR (`62bf2e77`) deliberately removed a string-prefix-based catch-rewrap that this PR introduced and then proved was brittle — the new shape is the consolidation, not an inconsistency.

deanmarano:
Review #2 addressed:

**Comments 1 & 2** (UnsupportedVisitError vs Rails TypeError): the old PR description claimed we throw `TypeError`, the implementation throws `UnsupportedVisitError`. The implementation is correct (and the existing test `to-sql.test.ts:361` pins it as the public surface). The PR description has been updated to match, and the JSDoc on the error class explains the deliberate translation: Rails raises `TypeError("Cannot visit #{class}")` ([`activerecord/lib/arel/visitors/visitor.rb:36`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/visitor.rb#L36)); Trails raises a named `UnsupportedVisitError extends ArelError` for the same condition.

While addressing this I also moved `UnsupportedVisitError` and `NotImplementedError` into `packages/arel/src/errors.ts` so every arel-thrown error now extends `ArelError` — a single `instanceof` root for callers (commit 34c2bc04).

**Comment 3** (compile-time check on dispatch method names): TS `keyof T` only sees public members and the visit methods are mostly `protected`, so a static `keyof ToSql` constraint isn't expressible. Added a runtime assertion in the registration helper instead — typos/renames now throw `Error("ToSql dispatch: method 'X' is not defined on the prototype")` the moment `to-sql.ts` is imported, which is when the static block runs. Same failure surface as a compile error, just one tick later.

deanmarano:
Review #3 addressed (commit e0d47e25):

**Comment 1 — `NodeCtor` typed with `never[]` instead of `unknown[]`:** kept the `never[]` signature and added a JSDoc explaining why. TS is contravariant in constructor parameter types, so `never[]` is the *only* signature that accepts ctors of arbitrary arity — `Binary(left, right)`, `BoundSqlLiteral(sql, binds)`, etc. I tried the suggested `unknown[]` and TS rejected every multi-arg ctor with TS2345 (e.g. `Type 'new (relation?: Node | null) => UpdateStatement' is not assignable to type 'abstract new (...args: unknown[]) => Node'`). `never[]` is the right type for an opaque map key — the dispatch cache never constructs nodes — and the doc now states that explicitly.

**Comment 2 — distinguish 'unknown node type' from 'mis-registered handler':** good catch. Split the two error paths in `Visitor.visit`:
- No entry in the dispatch cache → `Unknown node type: X` (the genuine missing-handler case).
- Entry in the cache but the method isn't defined on the instance → `Dispatch method 'X' is not defined on Y for node Z` (always a registration bug).

Added a test pinning the new branch (`distinguishes a mis-registered method from an unknown node type`).

deanmarano:
Review #4 addressed (commit 40eda8db): the memoization test now uses a fresh `Visitor` subclass it owns, with a pre-assertion that `dispatchCache().has(B)` is `false` before the `accept` call. The test now actually proves memoization end-to-end rather than reading state another test may have populated.

deanmarano:
Review #5:

**Comment 1 — re-export `UnsupportedVisitError`/`NotImplementedError` from `to-sql.ts` for back-compat:** declining. The author explicitly directed me to drop the re-exports and fix importers when these classes were moved into `errors.ts` (which is why commit 34c2bc04 was structured that way). The public surface is `packages/arel/src/visitors/index.ts` (which now re-exports from `../errors.js`) and the package barrel — anyone reaching into `visitors/to-sql.js` directly is consuming an internal path. Adding a back-compat re-export would re-introduce exactly the structural duplication we just removed.

**Comments 2 & 3 — use `NodeVisitor<T>` instead of inline `{ visit(n: Node): T }`:** good catch, fixed in commit 35b8b444. Imports `NodeVisitor` from `../nodes/node.js` and uses it in the test-only `A`/`C` Node subclasses. The dispatch tests now break loudly if the public visitor interface changes.

PR #994, opened 2026-04-29
feat(activemodel): port Clusivity privates (delimiter, isInclude, inclusionMethod)

deanmarano:
## Summary
Track A1 of [docs/activemodel-privates-100-plan.md](../blob/main/docs/activemodel-privates-100-plan.md). Mirrors `ActiveModel::Validations::Clusivity` (Rails `clusivity.rb`) verbatim:

- `ERROR_MESSAGE` — the canonical Rails error string.
- `delimiter()` — `clusivity.rb:31-33`. Memoized per validator instance via `_delimiterCache` with explicit Ruby-`||=` semantics (re-evaluates on `null`/`undefined`/`false`, not JS truthiness — `0` and `""` cache correctly). The fallback uses Ruby-`||` semantics too, so `options.in === false` falls through to `options.within`.
- `inclusionMethod(enumerable)` — `clusivity.rb:40-50`. Returns `"include?"` today; the Range cover-vs-include branch slots in cleanly when a TS Range type lands.
- `isInclude(record, value)` — `clusivity.rb:21-29`. Calls `this.resolveValue(record, this.delimiter())` — closes the `resolveValue` consumption gap flagged for Clusivity in PR #971. Routes through `this.delimiter()` / `this.inclusionMethod(...)` so subclass overrides are honored (Rails Ruby method-lookup semantics).
- `checkValidityBang()` — `clusivity.rb:14-18`. Symmetric with `isMemberOf`: accepts string (substring), Set/Map (`.has`), Array, duck-typed `.includes`/`.has`, generic iterables, and callables.
- `isMemberOf` throws `TypeError` when the resolved members value is `null`/`undefined` (matches Rails' loud `nil.include?` `NoMethodError` instead of silently returning `false`).
- Shared `exceptInWithinMergeValue(options, value)` helper for the Rails error-payload shape.

`InclusionValidator` and `ExclusionValidator` attach the four privates to their **prototypes** (not class fields) so the helpers are present during `EachValidator`'s constructor-time `checkValidity()` call. Class fields would be `undefined` at that point — caused a CI failure that surfaced and was fixed in commit `84fa7836`. They route their dispatch through `this.isInclude(record, value)` per `inclusion.rb:11` / `exclusion.rb:11`. Error payload uses `options.except(:in, :within).merge!(value: value)`.

Drops the trails-specific top-level helpers (`checkClusivityValidity`, `isMember`, `isExcluded`) — only `inclusion.ts` and `exclusion.ts` consumed them, and both now go through the Clusivity mixin.

## Impact
- `pnpm api:compare --privates --package activemodel`:
  - `clusivity.rb` 2/5 → **5/5 (100%)**
  - `inclusion.rb` 3/6 → **6/6 (100%)**
  - `exclusion.rb` 3/6 → **6/6 (100%)**
  - Overall +13 privates closed: 454/607 (74.8%) → **467/607 (76.9%)**.
- `pnpm api:compare --package activemodel`: stays **433/433 (100%)** — no public regression.
- `pnpm test:compare --package activemodel`: stays 959/963 (no change).

## Test plan
- [x] `pnpm vitest run packages/activemodel` — 1592 / 1592 passing
- [x] `pnpm api:compare --package activemodel` — 433/433
- [x] `pnpm api:compare --privates --package activemodel` — 467/607 (+13)
- [x] `pnpm test:compare --package activemodel` — 959/963 (0)

## Copilot review history (9 rounds, converged round 9)

Round 9 produced no new comments — review loop converged. Resolved over the prior rounds:

- **Round 1** — `isMemberOf` Rails-faithful for String (substring) / Map (key membership) / duck-typed `.includes`/`.has` collections; header comment matched actual attachment shape; dropped inverted `allowNil` pre-skip in inclusion/exclusion (rely on `EachValidator`'s `allow_nil` dispatch); updated `validates inclusion of with allow nil` test body to mirror `inclusion_validation_test.rb:100-106` (`allow_nil: true`); extracted shared `exceptInWithinMergeValue` helper.
- **Round 2** — `delimiter` memo only treats truthy cache as a hit (precursor); `isInclude` routes through `inclusionMethod(members)` and `testMembership` so the cover-vs-include branch slots in cleanly.
- **Round 3** — `delimiter` memo uses explicit `!== undefined && !== null && !== false` (Ruby `||=` semantics — `0` and `""` are truthy in Ruby); `checkValidityBang` accepts the same shapes `isMemberOf` does (symmetric duck-type).
- **Round 4** — `delimiter` fallback uses Ruby `||` semantics (`undefined`/`null`/`false` → fallback) instead of JS `??`, so `options.in === false` falls through to `options.within`.
- **Round 5** — Routed `isInclude` and `checkValidityBang` through `this.delimiter()` / `this.inclusionMethod(...)` so subclass overrides are honored.
- **Round 6** — Same flag as a CI failure (3 comments, all the bootstrapping bug). **Fixed via prototype attachment** — switching from `delimiter = delimiter` class fields to `InclusionValidator.prototype.delimiter = delimiter` so helpers are available during `super()` / constructor-time `checkValidity()`. `declare` typings keep public type contract.
- **Round 7** — Header comment matches prototype attachment; collapsed duplicate `typeof === "string"` checks in `checkValidityBang`.
- **Round 8** — `isMemberOf` throws `TypeError` on `null`/`undefined` members (Rails `nil.include?` raises `NoMethodError`).
- **Round 9** — No new comments — converged.

Ready for human review.

### DIFF
--- a/packages/activemodel/src/validations/clusivity.ts
+++ b/packages/activemodel/src/validations/clusivity.ts
@@ -44,7 +44,10 @@ interface ClusivityHost {
  * per validator instance, matching Rails' `||=` semantics.
  */
 export function delimiter(this: ClusivityHost): unknown {
-  if (this._delimiterCache !== undefined) return this._delimiterCache;
+  // Rails `@delimiter ||= ...` re-evaluates if the cached value is
+  // nil/false. Mirror that — only treat a truthy cache as a hit so
+  // null / false / "" recompute on the next call.
+  if (this._delimiterCache) return this._delimiterCache;
   this._delimiterCache = this.options.in ?? this.options.within;
   return this._delimiterCache;
 }
@@ -88,8 +91,21 @@ export function inclusionMethod(_enumerable: unknown): "include?" | "cover?" {
  */
 export function isInclude(this: ClusivityHost, record: unknown, value: unknown): boolean {
   const members = this.resolveValue(record, delimiter.call(this));
+  // Rails: `members.public_send(inclusion_method(members), v)`. Route
+  // through inclusionMethod so the cover-vs-include branch slots in
+  // when a Range type lands without reworking the call path.
+  const method = inclusionMethod(members);
   if (Array.isArray(value)) {
-    return value.every((v) => isMemberOf(members, v));
+    return value.every((v) => testMembership(members, v, method));
+  }
+  return testMembership(members, value, method);
+}
+
+function testMembership(members: unknown, value: unknown, method: "include?" | "cover?"): boolean {
+  if (method === "cover?") {
+    // Range#cover? semantics — start/end endpoint check. No first-class
+    // Range type in TS yet; if/when it lands, dispatch goes here.
+    return isMemberOf(members, value);
   }
   return isMemberOf(members, value);
 }
--- a/packages/activemodel/src/validations/clusivity.ts
+++ b/packages/activemodel/src/validations/clusivity.ts
@@ -55,7 +55,13 @@ export function delimiter(this: ClusivityHost): unknown {
   ) {
     return this._delimiterCache;
   }
-  this._delimiterCache = this.options.in ?? this.options.within;
+  // Rails: `options[:in] || options[:within]` — Ruby's `||` falls
+  // back when the left side is nil OR false, not just nullish. JS `??`
+  // would cache an explicit `false` and ignore `within`; reproduce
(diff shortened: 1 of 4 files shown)


deanmarano:
Copilot review loop converged at round 9 (zero new comments). PR body updated with the full 9-round resolution history. Ready for human review.

PR #998, opened 2026-04-29
refactor(activerecord): replace incidental new Date with Temporal

deanmarano:
## Summary

PR 11a of the Temporal migration's incidental-Date audit. Three isolated places in activerecord used `new Date()` for timestamp formatting where `Temporal` is the canonical source:

- `migration.ts#nextMigrationNumber` — version digits sourced from `Temporal.Now.instant().toString()`.
- `internal-metadata.ts#currentTime` — schema-migrations `updated_at` string sourced from `Temporal.Now.instant().toString({ smallestUnit: "millisecond" })`.
- `model-codegen.ts` — `now?: Date` widened to `now?: Temporal.Instant`. Dump-header timestamp uses `now.toString({ smallestUnit: "millisecond" })` so the output stays byte-identical to the prior `Date#toISOString` form.

The remaining `new Date` / `instanceof Date` hits in `activerecord/src` (attribute-inspection, base, query-methods value-equality, relation cache-key paths, time-zone-conversion) all handle externally-supplied values where Date interop is intentional. Those land in **PR 11b** as `// boundary:` annotations.

## Test plan

- [x] `vitest run packages/activerecord/src/{model-codegen,internal-metadata,migration}.test.ts` — 180/180 pass
- [x] `vitest run packages/activerecord/src/bin/trails-models-dump.test.ts` — 13/13 pass
- [x] `vitest run packages/activerecord` — 9915/9915 pass (full suite)
- [x] `pnpm typecheck`
- [x] Diff size: 24 LOC under the 300 ceiling


deanmarano:
Addressed in b2fd0148:

- **#1, #2** — fixed. `roundingMode: 'trunc'` passed to `Temporal.Instant#toString` in both model-codegen and internal-metadata. Matches prior `Date#toISOString` truncation semantics exactly.
- **#3** — fixed. Comment notation now uses `HH:mm:ss.SSS`.
- **#4 (widen `now` to `Date | Temporal.Instant`)** — won't fix. `model-codegen` is internal trails infrastructure (powers the `trails-models-dump` CLI tool), not a stable public API. The migration plan's 'no Date interop' decision intentionally narrows internal APIs to Temporal types — adding a Date overload would defeat the lint rule landing in PR 12. End-users of `trails-models-dump` interact via the CLI, which doesn't accept a `now` arg.

PR #1000, opened 2026-04-29
fix(runner): clear stale .runner before config.sh on container restart

deanmarano:
## Summary

- 12 of 16 dokku `gh-runner` replicas were crash-looping with `Cannot configure the runner because it is already configured`. Pool was running at ~25% capacity for ~21 hours.
- Root cause: when a container exits uncleanly (host reboot, OOM, SIGKILL), `.runner` is left in the writable layer; on restart, `./config.sh --unattended` refuses and exits 1 forever.
- Commit 1: detect `.runner` at entrypoint start and clear local state files before invoking `config.sh`.
- Commit 2: also DELETE any GitHub-side records registered under this host's prefix. The existing EXIT trap only deletes records under *this* boot's `\$RUNNER_NAME`, so without the sweep a record leaks each time a container recovers from an unclean exit (proven by the orphan we manually deleted post-deploy).

## Test plan

- [x] Deployed to dokku twice (commit 1, then commit 2). Both deploys produced 16/16 healthy containers and 16 registered GH runners after rolling restart.
- [x] Manually deleted the leaked orphan record post-commit-1 (validating the gap that commit 2 closes).
- [x] Static review of new branch: jq filter scopes to `\$SAFE_HOST-` prefix only, recovery path is idempotent and pipefail-safe.

### DIFF
--- a/infra/runner/entrypoint.sh
+++ b/infra/runner/entrypoint.sh
@@ -66,6 +66,17 @@ cleanup() {
 }
 trap cleanup EXIT
 
+# Clear stale config from a prior unclean exit (host reboot, OOM, SIGKILL).
+# Without this, a leftover `.runner` from the previous boot makes config.sh
+# refuse with "Cannot configure the runner because it is already configured"
+# and the container crash-loops forever (exit 1 → restart → same error).
+# The orphan GitHub-side record (if any) is reaped by the EXIT trap's
+# cleanup-by-name lookup, or by the next ephemeral cycle that reuses the host.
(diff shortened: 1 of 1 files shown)


deanmarano:
Already addressed in commit 035fc77a (pushed before this review fired). The recovery path now sweeps any GitHub records matching the `$SAFE_HOST-` prefix before registering — the same gap Copilot flagged. The misleading comment from commit 1 was rewritten to acknowledge that the EXIT trap doesn't cover the prior boot's name.

PR #1003, opened 2026-04-29
refactor(arel): port to_sql.rb statement visitors + 5 private helpers

deanmarano:
## Summary

PR 2 in the path to 100% on `pnpm api:compare --package arel --privates`. Mirrors the first 100 lines of `activerecord/lib/arel/visitors/to_sql.rb` plus the named-quoting/comment helpers from the bottom of the file.

Builds on #993 — the dispatch infrastructure landed there means renaming a visit method is a one-line dispatch-table edit, not a code-path change.

### Statement visit-method renames (project camelCase form of Rails' `visit_Arel_Nodes_*`)

Each Trails method is named with the camelCased form of the Rails method (`visit_Arel_Nodes_X` → `visitArelNodesX`). Project style is camelCase across the codebase, so we don't leak literal Ruby snake_case into TS — instead `scripts/api-compare/conventions.ts#snakeToCamel` is extended to handle `_<Capital>` so api:compare matches the Ruby name to the TS method automatically.

| Trails before | Trails now | Rails source |
|---|---|---|
| `visitSelectStatement` | `visitArelNodesSelectStatement` | to_sql.rb:120 |
| `visitSelectCore` | `visitArelNodesSelectCore` | to_sql.rb:149 |
| `visitInsertStatement` | `visitArelNodesInsertStatement` | to_sql.rb:53 |
| `visitUpdateStatement` | `visitArelNodesUpdateStatement` | to_sql.rb:40 |
| `visitDeleteStatement` | `visitArelNodesDeleteStatement` | to_sql.rb:22 |
| `visitExists` | `visitArelNodesExists` | to_sql.rb:225 |
| `visitComment` | `visitArelNodesComment` | to_sql.rb:175 |
| `visitTrue` | `visitArelNodesTrue` | to_sql.rb |
| `visitFalse` | `visitArelNodesFalse` | to_sql.rb |
| `visitCasted` | `visitArelNodesCasted` | to_sql.rb |
| `visitValuesList` | `visitArelNodesValuesList` | to_sql.rb |

`mysql.ts` / `sqlite.ts` / `postgresql.ts` overrides updated to track. `visitArelNodesComment` is also rewritten to delegate sanitization to the new `sanitizeAsSqlComment` helper and to mirror Rails' shape exactly (no leading space; callers add it via `maybeVisit`).

### New visit method

`visitArelNodesOptimizerHints` (to_sql.rb:170). The OptimizerHints node carries hints in a typed `hints` field (`ReadonlyArray<string | SqlLiteral>`); the visitor sanitizes each and emits `/*+ ... */`.

### Private helpers (new)

| Helper | Rails source | Behavior |
|---|---|---|
| `collectNodesFor` | to_sql.rb:179 | spacer + injectJoin (no-op if empty) |
| `injectJoin` | to_sql.rb:897 | (replaces Trails-specific `visitArray` across 22 callers) |
| `maybeVisit` | to_sql.rb:891 | leading-space + visit if non-null |
| `collectOptimizerHints` | to_sql.rb:887 | adapted to Trails' inline `string[]` storage |
| `sanitizeAsSqlComment` | to_sql.rb:882 | SqlLiteral pass-through; strip newlines and `/* */` |
| `quoteTableName` | to_sql.rb:872 | SqlLiteral pass-through; default double-quoted with `.`-segment splitting (handles `schema.table`); embedded `"` doubled |
| `quoteColumnName` | to_sql.rb:877 | SqlLiteral pass-through; default double-quoted |

`visitTable` / `visitAttribute` / `visitUnqualifiedColumn` now route through `quoteTableName` / `quoteColumnName` instead of inline `replace(/"/g, '""')` so the schema-qualified form is consistent (FROM clause and column refs both emit `"schema"."table"."col"`).

`visitArelNodesSelectCore` refactored to use `collectNodesFor` / `maybeVisit` line-for-line with to_sql.rb:149.

Note: the MySQL backtick override for `quoteTableName`/`quoteColumnName` is the deferred follow-up tracked by the `arel MySQL identifier quoting` memory item.

## Coverage

| | Before | After |
|---|---|---|
| Overall | 638/820 (77.8%) | **662/820 (80.7%)** |
| `visitors/to_sql.rb` | 7/118 (6%) | **26/118 (22%)** |

(Inheritance `68/69` is pre-existing on main — `UnsupportedVisitError` was moved into `errors.ts` by #993; not introduced here.)

## Tests

- `sanitizeAsSqlComment` SqlLiteral pass-through and delimiter/newline stripping.
- `quoteTableName` / `quoteColumnName` for bare names, schema-qualified names, embedded `"` doubling, and SqlLiteral pass-through.
- `visitArelNodesOptimizerHints` emits `/*+ ... */` for an array of hints (string and SqlLiteral), with comment-delimiter stripping.
- `collectNodesFor` spacer+connector behavior on multi-WHERE; no-op on empty lists.
- Comment placement: regression test pinning single-space `SELECT … /* hi */ /* bye */` shape across SelectStatement and SelectCore.
- Schema-qualified `Table("schema.table")` round-trip.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1105/1105 passing
- [x] `pnpm tsx scripts/api-compare/compare.ts --package arel --privates` — 662/820, to_sql.rb 26/118

### DIFF
--- a/packages/arel/src/nodes/unary.ts
+++ b/packages/arel/src/nodes/unary.ts
@@ -81,19 +81,18 @@ export class GroupingSet extends GroupingElement {
 
 export class Group extends Unary {}
 /**
- * `OptimizerHints#expr` is a list of hint strings (or SqlLiteral) — Rails'
- * Arel::Nodes::OptimizerHints stores `[hint1, hint2, ...]`. The Unary base
- * types `expr` as `Node | string | number | null` and TS won't let us
- * widen to an array via `declare`, so the constructor accepts the array
- * and we expose it via a typed `hints` getter. The visitor uses `hints`.
+ * Rails' `Arel::Nodes::OptimizerHints` stores `[hint1, hint2, ...]` and the
+ * visitor iterates them. The Unary base types `expr` as
+ * `Node | string | number | null` — TS won't let us widen `expr` to an array
+ * via `declare`, so the hints live in their own typed field and `expr` is
+ * left as `null` (consistent with the declared base type).
  */
 export class OptimizerHints extends Unary {
-  constructor(hints: ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>) {
-    super(hints as unknown as null);
-  }
+  readonly hints: ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>;
 
-  get hints(): ReadonlyArray<string | import("./sql-literal.js").SqlLiteral> {
-    return this.expr as unknown as ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>;
+  constructor(hints: ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>) {
+    super(null);
+    this.hints = hints;
   }
 }
 export class RollUp extends Rollup {}
--- a/packages/arel/src/nodes/unary.ts
+++ b/packages/arel/src/nodes/unary.ts
@@ -80,5 +80,20 @@ export class GroupingSet extends GroupingElement {
 }
 
 export class Group extends Unary {}
-export class OptimizerHints extends Unary {}
+/**
+ * `OptimizerHints#expr` is a list of hint strings (or SqlLiteral) — Rails'
+ * Arel::Nodes::OptimizerHints stores `[hint1, hint2, ...]`. The Unary base
+ * types `expr` as `Node | string | number | null` and TS won't let us
+ * widen to an array via `declare`, so the constructor accepts the array
+ * and we expose it via a typed `hints` getter. The visitor uses `hints`.
+ */
+export class OptimizerHints extends Unary {
+  constructor(hints: ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>) {
+    super(hints as unknown as null);
+  }
+
+  get hints(): ReadonlyArray<string | import("./sql-literal.js").SqlLiteral> {
+    return this.expr as unknown as ReadonlyArray<string | import("./sql-literal.js").SqlLiteral>;
+  }
+}
 export class RollUp extends Rollup {}
--- a/packages/arel/src/visitors/mysql.ts
+++ b/packages/arel/src/visitors/mysql.ts
@@ -43,9 +43,7 @@ export class MySQL extends ToSql {
       this.visit(node.lock);
     }
 
-    if (node.comment) {
-      this.visit(node.comment);
-    }
+    this.maybeVisit(node.comment ?? null);
 
     return this.collector;
   }
--- a/packages/arel/src/visitors/mysql.ts
+++ b/packages/arel/src/visitors/mysql.ts
@@ -9,7 +9,7 @@ import { ToSql } from "./to-sql.js";
  * Mirrors: Arel::Visitors::MySQL
  */
 export class MySQL extends ToSql {
-  protected override visitSelectStatement(node: Nodes.SelectStatement): SQLString {
+  protected override visit_Arel_Nodes_SelectStatement(node: Nodes.SelectStatement): SQLString {
     if (node.with) {
       this.visit(node.with);
       this.collector.append(" ");
@@ -50,7 +50,7 @@ export class MySQL extends ToSql {
     return this.collector;
   }
 
-  protected override visitSelectCore(node: Nodes.SelectCore): SQLString {
+  protected override visit_Arel_Nodes_SelectCore(node: Nodes.SelectCore): SQLString {
     this.collector.append("SELECT");
 
     this.emitOptimizerHints(node);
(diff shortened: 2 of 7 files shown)


deanmarano:
Review #1 addressed (commit 12ee4ab6):

**Comment 1 — schema-qualified table identifier mismatch:** real bug, fixed. Moved the dot-split logic into `quoteTableName` itself so both `visitTable` and `visitAttribute` go through the same helper. `new Table("schema.table").project(t.get("col")).toSql()` now emits `SELECT "schema"."table"."col" FROM "schema"."table"` consistently. Mirrors Rails' adapter-side behavior (e.g. PG's [`quote_table_name`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb) splits on the schema separator and quotes each segment). Test added pinning the round-trip.

**Comment 2 — `OptimizerHints.expr` typing:** went with the typed-narrowing path. `Unary.expr` is `Node | string | number | null`, and TS rejects `declare readonly expr: (string|SqlLiteral)[]` because the array isn't a subtype of the base union. Added a constructor that takes the array and exposed a typed `hints` getter (`ReadonlyArray<string | SqlLiteral>`). The visitor now reads `node.hints` rather than casting `node.expr`. Tests no longer use `as unknown as Nodes.Node` and a SqlLiteral pass-through test was added.

deanmarano:
Review #2 addressed (commit b6f0d718):

**Comment 1 — double-space before SELECT comments** (real bug). `visitArelNodesComment` was prepending a leading space per value, and SelectCore's `maybeVisit(node.comment)` was adding *another*. SelectStatement bypassed `maybeVisit` and called `visit` directly, which is why the test suite didn't catch it. Now mirrors [`to_sql.rb:175`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/to_sql.rb#L175) exactly: the visitor emits `/* ... */` blocks joined by a single space with no leading separator; all callers (SelectStatement, SelectCore, MySQL + SQLite SelectStatement overrides) now route through `maybeVisit` for the leading space. Regression test added pinning `SELECT … /* hi */ /* bye */` shape.

**Comment 2 — OptimizerHints typing inconsistency**: switched to a dedicated typed `hints` field. `Unary.expr` is now left as `null` (consistent with its declared `Node | string | number | null` type), and the array lives in `OptimizerHints#hints: ReadonlyArray<string | SqlLiteral>`. No more runtime/static disagreement; the visitor's read of `node.hints` is unchanged.

deanmarano:
Review #3 addressed:

**Comments 1 & 2 — use Ruby method names in Rails-reference comments** (commit 721ee28b): right call. The mass-camelCasing also caught comment text where I was citing the Rails source by name. Comments now use the literal Ruby form (`visit_Arel_Nodes_Cte`, `visit_Arel_Nodes_GroupingElement`, etc.) so the reference is greppable in the Rails repo. Method declarations themselves stay in project camelCase (`visitArelNodesX`) — only the prose changed.

**Comment 3 — PR description vs implementation mismatch**: updated. The description now matches reality: methods are camelCase `visitArelNodesX` (project style), and api:compare matches them to Ruby's `visit_Arel_Nodes_X` via the `snakeToCamel` extension in `scripts/api-compare/conventions.ts`. The earlier draft text claiming literal Rails-snake_case identifiers in TS reflected an interim state we walked back at the user's direction.

deanmarano:
Review #4 — declining the change.

The reviewer's suggestion is to filter empty sanitized hints and no-op when the list collapses, matching Trails' existing `emitOptimizerHints` behavior. But `emitOptimizerHints` is the **inline-string-array** path (Trails' `SelectCore.optimizerHints: string[]`), where filtering noise out of user input is reasonable.

`visitArelNodesOptimizerHints` is the visitor for the explicit `Arel::Nodes::OptimizerHints` node, and Rails' [`to_sql.rb:170-173`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/to_sql.rb#L170-L173) is unambiguous — no filtering, no empty guard:

```ruby
def visit_Arel_Nodes_OptimizerHints(o, collector)
  hints = o.expr.map { |v| sanitize_as_sql_comment(v) }.join(" ")
  collector << "/*+ #{hints} */"
end
```

Rails will emit `/*+  */` for an empty hint array — the user is responsible for not constructing a degenerate node. Trails matches that. Filtering here would mask the case where a caller misconstructed the node (e.g. `new OptimizerHints([])`) instead of letting the bad SQL surface, and would diverge from the upstream method we're mirroring.

Rails fidelity is the stated objective, so I'm keeping the visitor faithful to the Rails source. `emitOptimizerHints` (different code path, non-Rails-mirrored) keeps its filtering because it's an internal helper for the inline-array storage.

PR #1008, opened 2026-04-29
refactor(arel): camelCase the to_sql.rb visit-method names + identifier-quoting helper routing

deanmarano:
## Summary

PR 3 in the arel privates push. Two phases:

1. **Bulk rename** of 72 visit methods to the camelCased form of Rails' literal `visit_Arel_Nodes_X` names (continuing the convention from #1003 — `scripts/api-compare/conventions.ts#snakeToCamel` handles the Ruby↔TS bridge).
2. **Helper-routing fix** at four sites that were still interpolating identifiers directly with `` `"${name}"` ``, bypassing the `quoteTableName` / `quoteColumnName` helpers PR #1003 introduced. The bypass produced invalid SQL on identifiers containing a literal `"` and skipped any dialect-specific override.

### Visit-method renames (72 methods, 2 commits for reviewability)

**Commit 1 (40de3d75)** — window/order/aggregate/CTE batch (38 methods):

| Category | Methods |
|---|---|
| Set ops | Union, UnionAll, Intersect, Except |
| CTE | With, WithRecursive, TableAlias, Cte |
| Window | Over, NamedWindow, Window, Rows, Range, Preceding, Following, CurrentRow |
| Order/limit/lock | Offset, Limit, Lock, Ascending, Descending, NullsFirst, NullsLast |
| Grouping | Grouping, HomogeneousIn, Group |
| Functions | NamedFunction, Extract, Count, Sum, Max, Min, Avg |
| Misc | Distinct, DistinctOn, Bin, Filter, Fragments |

**Commit 2 (d14a18fa)** — predicate/join/boolean batch (34 methods):

| Category | Methods |
|---|---|
| Joins | JoinSource, InnerJoin, OuterJoin, RightOuterJoin, FullOuterJoin, StringJoin, On |
| Predicates | Equality, NotEqual, IsDistinctFrom, IsNotDistinctFrom, In, NotIn, Between, Assignment, As, Regexp, NotRegexp, Matches, DoesNotMatch, UnqualifiedColumn |
| Boolean | And, Or, Not |
| Case | Case, When, Else |
| Literals/binds | BindParam, SqlLiteral, BoundSqlLiteral, InfixOperation, UnaryOperation |
| Table/Attribute | `visitTable` → `visitArelTable` (Rails: `visit_Arel_Table` — Table isn't under the Nodes module); `visitAttribute` → `visitArelAttributesAttribute` (Attribute is under `Arel::Attributes::Attribute`) |

Dispatch table entries and dialect overrides (`mysql.ts`, `sqlite.ts`, `postgresql.ts`) updated to track. The dispatch helper's runtime `ToSql.prototype[m]` guard catches any rename misses immediately at module-import time.

### Identifier-quoting bug fix (commit 41b6943b)

Four sites still interpolated identifiers directly via `` `"${name}"` ``, which:
- Did not escape embedded double-quotes (so a name like `we"ird` produced unbalanced SQL).
- Bypassed dialect overrides (e.g. the deferred MySQL backtick-quoting follow-up tracked by the `arel MySQL identifier quoting` memory item would never reach these sites).

| Site | Helper now used | Rails parity |
|---|---|---|
| `visitArelNodesAssignment` (UPDATE SET column) | `quoteColumnName` | Rails `visit_Arel_Nodes_Assignment` uses `quote_column_name` |
| `visitArelNodesNamedWindow` (window name) | `quoteColumnName` | Rails `visit_Arel_Nodes_NamedWindow` uses `quote_column_name` |
| `visitArelNodesTableAlias` (alias on a non-Grouping relation) | `quoteTableName` | Rails `visit_Arel_Nodes_TableAlias` uses `quote_table_name` |
| `visitArelNodesCte` (CTE name) | `quoteTableName` | Rails MySQL/Postgres CTE visitors call `quote_table_name` |

Regression test added: round-trips identifiers containing a literal `"` through UPDATE SET and CTE shapes.

## Coverage

| | Before | After |
|---|---|---|
| Overall | 662/820 (80.7%) | **751/820 (91.6%)** |
| `visitors/to_sql.rb` | 26/118 (22%) | **96/118 (81%)** |

(+89 hits — slightly more than the 72 renames because some flow through dialect-override matches too.)

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1107/1107 passing
- [x] `pnpm tsx scripts/api-compare/compare.ts --package arel --privates` — 751/820, to_sql.rb 96/118

### DIFF
--- a/packages/arel/src/visitors/mysql.ts
+++ b/packages/arel/src/visitors/mysql.ts
@@ -125,7 +125,7 @@ export class MySQL extends ToSql {
   // unlike the base ToSql which special-cases the bare name. The
   // relation prefix this leaves on for an Attribute is fine: MySQL's
   // `UPDATE t SET x = t.x + 1` is valid.
-  protected override visitUnqualifiedColumn(node: Nodes.UnqualifiedColumn): SQLString {
+  protected override visitArelNodesUnqualifiedColumn(node: Nodes.UnqualifiedColumn): SQLString {
     if (node.expr instanceof Node) {
       this.visit(node.expr);
     } else if (node.expr !== null) {
@@ -137,14 +137,14 @@ export class MySQL extends ToSql {
   // MySQL's null-safe equality is `<=>`. The standard `IS [NOT] DISTINCT
   // FROM` is supported only on MySQL 8.0.14+; the operator form works
   // on every MySQL version.
-  protected override visitIsNotDistinctFrom(node: Nodes.IsNotDistinctFrom): SQLString {
+  protected override visitArelNodesIsNotDistinctFrom(node: Nodes.IsNotDistinctFrom): SQLString {
     this.visitNodeOrValue(node.left);
     this.collector.append(" <=> ");
     this.visitNodeOrValue(node.right);
     return this.collector;
   }
 
-  protected override visitIsDistinctFrom(node: Nodes.IsDistinctFrom): SQLString {
+  protected override visitArelNodesIsDistinctFrom(node: Nodes.IsDistinctFrom): SQLString {
     this.collector.append("NOT ");
     this.visitNodeOrValue(node.left);
     this.collector.append(" <=> ");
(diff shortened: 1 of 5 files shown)


deanmarano:
Review #1 addressed (commit 41b6943b): all three suggestions are real consistency bugs — direct `"${name}"` interpolation that bypasses the `quoteTableName` / `quoteColumnName` helpers introduced in #1003. Routing them through the helpers also closes the same hole at a fourth site Copilot didn't flag (`visitArelNodesTableAlias` when `relation` is not a `Grouping`).

Sites fixed:
- `visitArelNodesAssignment` — UPDATE SET column → `quoteColumnName`
- `visitArelNodesNamedWindow` — window name → `quoteColumnName` (matches Rails' `quote_column_name` for window aliases)
- `visitArelNodesTableAlias` — table alias when not wrapping a Grouping → `quoteTableName`
- `visitArelNodesCte` — CTE name → `quoteTableName` (matches Rails' `quote_table_name` for CTE identifiers)

Regression test added: round-trips identifiers containing a literal `"` through UPDATE SET and CTE shapes — both now emit `"co""l"` / `"w""in"` correctly.

deanmarano:
Review #2 addressed: title and description updated. The PR now explicitly calls out two phases — (1) the 72 rename commits (40de3d75 + d14a18fa) which are pure renames, and (2) the identifier-quoting helper-routing bug fix (41b6943b) which is the behavior-changing hunk. The fix was added in response to your review #1; updating the description here makes the dual scope clear for future readers and release notes.

PR #1019, opened 2026-04-29
feat(eslint): no-native-date rule enforcing the Temporal migration

deanmarano:
## Summary

PR 12 — final ESLint safety net for the Temporal migration. Adds `blazetrails/no-native-date` to flag JS `Date` *value* usage in domain code:

- `new Date(...)` and `new globalThis.Date(...)` / `window.Date` / `self.Date` / `global.Date` constructor calls.
- `x instanceof Date` (and the `globalThis.Date` etc. variants) runtime checks.

### Escape hatches

- **File-level** `@boundary-file:` JSDoc-shaped directive in the file *header* — exempts the entire file. Used by Date-typed Rails-port modules (`duration.ts`, `time-ext.ts`, `time-zone.ts`, etc.) and HTTP/RFC-spec boundary code (cookies, conditional-get, etc.).
- **Line-level** `boundary:` keyword in a comment that is either:
  - on the same line as the `Date` reference (anchored on `node.callee` for `new` and `node.right`/`.property` for `instanceof`, so multi-line `globalThis.Date` shapes carry the marker on the `.Date` line);
  - leading the enclosing top-level statement; or
  - inside the enclosing statement before the offending node (handles `} /* boundary */ else if` chains).

### Out of scope (intentional)

- `Date.{now,parse,UTC}` — return numbers, not propagating Date values.
- `: Date` type references — constrain flow without creating Date instances.

Locally-bound `Date` (e.g. `activerecord/src/type.ts` importing the AR `Type::Date` class) is naturally exempt via ESLint scope analysis.

### Wiring

Enabled at `error` for the seven Rails-mirroring packages plus `rack`/`trailties`/`website`. Allowlist:

- `packages/activesupport/src/temporal.ts` — canonical `Date`↔`Instant` adapter.
- `packages/activesupport/src/testing/**` and `testing-helpers.ts` — test infrastructure.

### Audit fallout

The rule found a handful of sites that PR 11's manual sweeps missed. Added boundary annotations to:

- `activerecord` connection-adapters (`abstract/quoting`, `database-statements`, `mysql/quoting`, `sqlite3/quoting`, `postgresql/oid/range`, `sqlite3-adapter`) — file-level `@boundary-file:` and line-level annotations.
- `activerecord/connection-adapters/abstract/transaction.ts` — `Notifications.Event` Date timestamp.
- `activerecord/relation.ts` and `relation/query-methods.ts` — relocated/added boundary markers.
- `actionpack/actioncontroller/base.ts` — `Temporal.Instant`→`Date` bridge in `freshWhen`.
- `arel/visitors/dot.ts` — added by PR #1020 (rebased in).

`pnpm lint` runs clean across the whole repo with the rule at `error`. Any regression PR fails CI.

### Review status

Copilot review cycles concluded (10 rounds). The final pass surfaced **no new comments** — all prior concerns were resolved across the iterations:

- File-header detection narrowed to header-only JSDoc-shaped block comments.
- Same-line marker check anchored on the `Date` reference (incl. `.Date` for member access).
- `globalThis.Date` / `window.Date` / etc. detection added with caught-and-annotated SQLite-adapter site.
- Line-level escape hatch tightened from "enclosing block" to "enclosing statement" with documented multi-line / `else if` ergonomics.
- Per-file `getAllComments()` cached once instead of per-node.
- Two NaN-Date guard sites and Invalid-Date inspect rendering hardened along the way.

Together with PRs 1–11, this **completes the Temporal migration**.

## Test plan

- [x] `node eslint/no-native-date.test.mjs` — rule unit tests pass (12 valid + 8 invalid scenarios)
- [x] `pnpm lint` — clean across the whole repo
- [x] `vitest run packages` — full test suite passes
- [x] `pnpm typecheck`

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -88,7 +88,7 @@ export default defineConfig(
     },
   },
 
-  // ── blazetrails plugin (shared by no-node-builtins + rails-private-jsdoc) ──
+  // ── blazetrails plugin (no-node-builtins + rails-private-jsdoc + no-native-date) ──
   // Registered without a `files` restriction so any block below can
   // reference its rules without re-declaring the plugin.
   {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -8,6 +8,7 @@ import vitest from "@vitest/eslint-plugin";
 import noNodeBuiltins from "./eslint/no-node-builtins.mjs";
 import noProcessBypass from "./eslint/no-process-bypass.mjs";
 import railsPrivateJsdoc from "./eslint/rails-private-jsdoc.mjs";
+import noNativeDate from "./eslint/no-native-date.mjs";
 
 export default defineConfig(
   {
@@ -97,6 +98,7 @@ export default defineConfig(
           "no-node-builtins": noNodeBuiltins,
           "no-process-bypass": noProcessBypass,
           "rails-private-jsdoc": railsPrivateJsdoc,
+          "no-native-date": noNativeDate,
         },
       },
     },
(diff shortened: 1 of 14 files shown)


deanmarano:
Already addressed in 1a1478e0b (just pushed before this review surfaced) — the doc comment was rewritten to match the implementation, and the implementation was simultaneously tightened so 'inside the enclosing block' is no longer accepted. The three forms now are: (a) same-line trailing, (b) leading the enclosing top-level statement, (c) inside the enclosing statement before the offending node.

PR #1020, opened 2026-04-29
refactor(arel): port dot.rb visitor to 100% — arel privates 820/820 ✓

deanmarano:
## Summary

**Final PR in the arel privates push.** With this merged, `pnpm api:compare --package arel --privates` reports **820/820 (100%)**.

Faithful Rails port of `Arel::Visitors::Dot` (`activerecord/lib/arel/visitors/dot.rb`). The previous Trails Dot was a generic reflection-based walk; this rewrite mirrors Rails' shape exactly so api:compare can match every method one-to-one.

## What changed

- **`DotNode` / `DotEdge`** classes match `Arel::Visitors::Dot::Node` and `::Edge` (Rails' Struct).
- **Internal state** (`nodes` / `edges` / `nodeStack` / `edgeStack` / `seen`) mirrors Rails' instance variables.
- **25 `visit_*` methods**, each declaring the named edges to descend — matches every method Rails enumerates in `dot.rb`.
- **Aliases** mirroring Rails' `alias`:
  - `visit_Regexp` → Regexp / NotRegexp
  - `visit_NoEdges` → CurrentRow / Distinct (and Trails extras: True / False / Quoted / BoundSqlLiteral / Fragments / SelectOptions)
  - `visit_Children` → And / Or / With / WithRecursive
  - `visitString` → SqlLiteral and primitive types via Trails' isPrimitive check (Rails aliases `visit_Time`/`visit_Date`/`visit_DateTime`/`visit_Integer`/`visit_BigDecimal`/`visit_Float`/`visit_Symbol`/`visit_TrueClass`/`visit_FalseClass`/`visit_NilClass`/`visit_Arel_Nodes_SqlLiteral`)
- **Core machinery** ports `visit_edge`, `edge`, `with_node`, `quote`, `to_dot` as `visitEdge`, `edge`, `withNode`, `quote`, `toDot` — same Rails-faithful behavior.
- **Dispatch table** registers all 25+ Rails-named visit methods. Trails-specific node types (joins, set ops, comparisons, ordering subclasses) inherit from registered ancestors (Unary / Binary / InfixOperation / Ordering / Function), so the `Visitor.resolveDispatch` prototype-walk fallback routes them at first use — no per-class registration needed.

## Trails-specific concessions

- Rails uses `object.object_id` for node IDs; Trails counts integers (works the same for identity in the seen-map).
- Hash / Array dispatch goes by JS type rather than the dispatch table (JS arrays and plain objects aren't constructible classes).
- `Quoted` / `True` / `False` / `BoundSqlLiteral` / `Fragments` / `SelectOptions` aren't enumerated in Rails' `dot.rb`; they're registered as `visit_NoEdges` (matches Rails' default-no-edges behavior for unenumerated nodes). Trails tests construct them.
- `Table` is registered lazily at first `accept()` call rather than in the static block, because at static-block-execution time the `Table` class is mid-load via a circular import through `tree-manager.js`. By the time the first instance is used, the class is fully resolved.

## Coverage

| | Before | After |
|---|---|---|
| Overall | 782/820 (95.4%) | **820/820 (100%) ✓** |
| `visitors/dot.rb` | 3/41 (7%) | **41/41 (100%) ✓** |

**Every visitor file is now at 100%.** Combined with the four prior PRs (#993, #1003, #1008, #1014), this completes the arel privates push from 635/820 (77.4%) → 820/820 (100%).

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1118/1118 passing (16 dot tests included)
- [x] `pnpm tsx scripts/api-compare/compare.ts --package arel --privates` — **820/820 (100%) ✓**

## Note on diff size

418+/54- in one file. Over the 300 LOC ceiling in CLAUDE.md, but this is a single-file faithful port of Rails' 299-line `dot.rb`. Splitting `Arel::Visitors::Dot` across PRs would leave intermediate states that don't compile or that emit broken graphs (the `visit` core depends on every visit method being present at first use). Keeping it as one cohesive port is the lower-risk path.

---

## Open follow-up nits (Copilot review #11, deferred — review-cycle cap reached)

Two further fidelity polish items surfaced after the cap:

1. **`visitString` of a Symbol** still uses `String(sym)` (yields `"Symbol(foo)"`) instead of Ruby's `Symbol#to_s` shape (`"foo"` via `sym.description ?? ""`). Carried over from review #10 nit 1; user direction was to skip it then.
2. **JS plain objects route through `visitHash` but `classNameOf` labels them `Object`** (their JS ctor name) rather than Rails' `Hash`. Adding a plain-object check in `classNameOf` to return `"Hash"` would match Rails dot output.

Neither affects the 820/820 coverage; both are output-shape polish.

### DIFF
--- a/packages/arel/src/visitors/dot.test.ts
+++ b/packages/arel/src/visitors/dot.test.ts
@@ -156,6 +156,30 @@ describe("TestDot", () => {
       expect(out).toContain('say \\"hi\\"');
     });
 
+    it("OptimizerHints renders its hints field (not Unary's null expr)", () => {
+      const node = new Nodes.OptimizerHints(["IDX(t1)", "MAX_EXEC_TIME(1000)"]);
+      const out = dot.compile(node);
+      expect(out).toContain("OptimizerHints");
+      expect(out).toMatch(/-> \d+ \[label="hints"\];/);
+      expect(out).toContain("IDX(t1)");
+      expect(out).toContain("MAX_EXEC_TIME(1000)");
+    });
+
+    it("non-Node bind values (ActiveModel::Attribute shape) don't crash", () => {
+      // Regression: Dot.visit used to call super.visit on any non-primitive
+      // non-array non-plain-object value, throwing UnsupportedVisitError
+      // on a class instance the dispatch table didn't know about.
+      const fakeAttribute = {
+        valueBeforeTypeCast: 42,
+      };
+      const bind = new Nodes.BindParam(fakeAttribute);
+      const out = dot.compile(bind);
+      expect(out).toContain("BindParam");
+      // visitActiveModelAttribute walks valueBeforeTypeCast.
+      expect(out).toMatch(/-> \d+ \[label="valueBeforeTypeCast"\];/);
+      expect(out).toContain("42");
+    });
+
     it("visitHash preserves both key and value (Rails parity)", () => {
       // Mirrors Rails dot.rb:227 — visit_Hash emits one edge per entry
       // labeled "pair_<i>" pointing at an Array node, which itself emits
--- a/packages/arel/src/visitors/dot.test.ts
+++ b/packages/arel/src/visitors/dot.test.ts
@@ -111,4 +111,49 @@ describe("TestDot", () => {
     const out = dot.compile(stmt);
     expect(out).toContain("DeleteStatement");
   });
+
+  describe("output structure (Rails parity)", () => {
+    it("emits the Rails dot.rb header and shape", () => {
+      const out = dot.compile(new Nodes.Distinct());
+      expect(out).toMatch(/^digraph "Arel" \{\n/);
+      expect(out).toContain("node [width=0.375,height=0.25,shape=record];");
+      expect(out).toMatch(/\n\}$/);
+      // A leaf node: id [label="<f0>Name"];
+      expect(out).toMatch(/^\d+ \[label="<f0>Distinct"\];$/m);
+    });
+
+    it("emits one edge per visit_edge declaration with the field name as label", () => {
+      // Binary -> left, right (two visit_edge calls).
+      const node = new Nodes.Equality(users.get("id"), new Nodes.SqlLiteral("1"));
+      const out = dot.compile(node);
+      expect(out).toMatch(/-> \d+ \[label="left"\];/);
+      expect(out).toMatch(/-> \d+ \[label="right"\];/);
+    });
+
+    it("emits an InfixOperation's three edges in Rails order: operator, left, right", () => {
+      const node = new Nodes.InfixOperation("+", users.get("age"), new Nodes.Quoted(1));
+      const out = dot.compile(node);
+      const operatorPos = out.indexOf('[label="operator"]');
+      const leftPos = out.indexOf('[label="left"]');
+      const rightPos = out.indexOf('[label="right"]');
+      expect(operatorPos).toBeGreaterThan(-1);
+      expect(operatorPos).toBeLessThan(leftPos);
+      expect(leftPos).toBeLessThan(rightPos);
+    });
+
+    it("collapses to a leaf for visit_NoEdges nodes (CurrentRow, Distinct)", () => {
+      const out = dot.compile(new Nodes.CurrentRow());
+      // Single node, no edges.
+      const edges = (out.match(/->/g) ?? []).length;
+      expect(edges).toBe(0);
+    });
+
+    it("escapes embedded double-quotes in side-field labels (quote helper)", () => {
+      const node = new Nodes.SqlLiteral('say "hi"');
+      const out = dot.compile(node);
+      // SqlLiteral is dispatched as visit_String — the value becomes a
+      // side-field on the parent node with quote() escaping the `"`.
+      expect(out).toContain('say \\"hi\\"');
+    });
+  });
 });
--- a/packages/arel/src/visitors/dot.test.ts
+++ b/packages/arel/src/visitors/dot.test.ts
@@ -173,6 +173,29 @@ describe("TestDot", () => {
       expect(out).not.toContain("undefined");
     });
 
+    it("Extract walks expr + field (Trails shape, not Rails' expressions + alias)", () => {
+      const node = new Nodes.Extract(users.get("created_at"), "year");
+      const out = dot.compile(node);
+      expect(out).toContain("Extract");
+      expect(out).toMatch(/-> \d+ \[label="expr"\];/);
+      expect(out).toMatch(/-> \d+ \[label="field"\];/);
+      expect(out).toContain("year");
+      // No edges to nil-shaped Rails fields.
+      expect(out).not.toMatch(/-> \d+ \[label="expressions"\];/);
+      expect(out).not.toMatch(/-> \d+ \[label="alias"\];/);
+    });
+
+    it("Exists walks expressions + alias (no spurious distinct edge)", () => {
+      const inner = new SelectManager(users).project(users.get("id")).ast;
+      const node = new Nodes.Exists(inner);
+      const out = dot.compile(node);
+      expect(out).toContain("Exists");
+      expect(out).toMatch(/-> \d+ \[label="expressions"\];/);
+      expect(out).toMatch(/-> \d+ \[label="alias"\];/);
+      // Generic Function visitor would have emitted a `distinct` edge.
+      expect(out).not.toMatch(/-> \d+ \[label="distinct"\];/);
+    });
+
     it("OptimizerHints renders its hints field (not Unary's null expr)", () => {
       const node = new Nodes.OptimizerHints(["IDX(t1)", "MAX_EXEC_TIME(1000)"]);
       const out = dot.compile(node);
--- a/packages/arel/src/visitors/dot.test.ts
+++ b/packages/arel/src/visitors/dot.test.ts
@@ -187,6 +187,48 @@ describe("TestDot", () => {
       );
     });
 
+    it("UpdateStatement walks groups and havings (Trails fields)", () => {
+      const stmt = new UpdateManager()
+        .table(users)
+        .set([[users.get("name"), "x"]])
+        .group(users.get("dept"))
+        .having(users.get("active").eq(true)).ast;
+      const out = dot.compile(stmt);
+      expect(out).toContain("UpdateStatement");
+      expect(out).toMatch(/-> \d+ \[label="groups"\];/);
+      expect(out).toMatch(/-> \d+ \[label="havings"\];/);
+    });
+
+    it("DeleteStatement walks groups and havings (Trails fields)", () => {
+      const stmt = new DeleteManager()
+        .from(users)
+        .group(users.get("dept"))
+        .having(users.get("active").eq(true)).ast;
+      const out = dot.compile(stmt);
+      expect(out).toContain("DeleteStatement");
+      expect(out).toMatch(/-> \d+ \[label="groups"\];/);
+      expect(out).toMatch(/-> \d+ \[label="havings"\];/);
+    });
+
+    it("repeated equal scalar primitives dedupe onto one DotNode (Rails singleton parity)", () => {
+      // Rails' true/false/Integers are singletons with stable object_id, so
+      // two visits of `true` reuse one node. Strings still distinct.
+      const v = new Visitors.Dot();
+      type Internals = { visit(o: unknown): void; toDot(): string };
+      v.compile(new Nodes.SqlLiteral("seed"));
+      (v as unknown as Internals).visit(true);
+      (v as unknown as Internals).visit(true);
+      (v as unknown as Internals).visit(42);
+      (v as unknown as Internals).visit(42);
+      const out = (v as unknown as Internals).toDot();
+      // Booleans and numbers each fire a single labeled node — repeats
+      // shouldn't allocate new ones.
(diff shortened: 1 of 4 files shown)


deanmarano:
Review #1 addressed (commit 03dea8782) — all three are real:

**Comments 1 & 3 — `edge()` and `withNode()` lacked try/finally.** Both methods push state onto a stack, run an arbitrary `block()`, and pop. If the block throws (e.g. `UnsupportedVisitError` from a node lacking a dispatch entry), the stack stays corrupted for the next `accept()` call. Wrapped each `block()` invocation in `try/finally` so the pop always runs — mirrors Ruby's `ensure`. Rails' [`dot.rb:248`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L248) and [`dot.rb:255`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L255) get this for free via Ruby's exception-safe blocks; we have to be explicit.

**Comment 2 — static `reg(Table, ...)` was dead code.** Right call. The earlier fix moved Table registration into `accept()`'s lazy block because the class is mid-load at static-block time via the circular import through `tree-manager.js`. The static entry registered against the partial module-shim ref and was overwritten on first `accept()` anyway. Removed; left a comment pointing readers at the lazy registration.

1123/1123 arel tests still passing.

deanmarano:
Review #2 addressed (commit 75baa291a):

**Comment 1 — `DotEdge.to` definite-assignment assertion was unsound.** Right; `to` is set later by `withNode()` once the destination Node exists, and stays unset for edges whose target was filtered out. Changed to `to?: DotNode` so the type matches the lifecycle, and replaced `toDot`'s `filter + e.to!` with a single `flatMap` truthy guard.

**Comment 2 — `visitHash` was losing the key.** Real bug. Rails [`dot.rb:227-229`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L227-L229):

```ruby
def visit_Hash(o)
  o.each_with_index do |pair, i|
    edge("pair_#{i}") { visit pair }
  end
end
```

The outer label is `pair_#{i}` (Rails-defined), and the body visits the *whole* `pair` — a 2-element [key, value] array — which dispatches to `visit_Array` and emits index-labeled inner edges for both halves. My old impl destructured the entry and only visited the value, dropping the key from the graph entirely.

Note that the reviewer's suggested fix (using the key as the outer edge label) would have lost Rails parity — Rails uses `pair_#{i}` for that label. The fix here visits the pair as Rails does, which preserves the key info via nested rendering rather than re-labeling.

Test added asserting both keys and both values appear in the graph; 1124/1124 arel tests passing.

deanmarano:
Review #3 addressed (commit dd0fbeeba) — both real bugs:

**Comment 1 — non-Node values crashed Dot.visit.** Right; my fall-through called `super.visit()` on any non-primitive non-array non-plain-object value, which dispatched and threw `UnsupportedVisitError` for class instances the dispatch didn't know about. The routing in `visit()` now branches:

- `instanceof Node || Table` → dispatch (Rails-named visitors)
- duck-typed ActiveModel::Attribute (has `valueBeforeTypeCast` *and* isn't a Node) → `visitActiveModelAttribute` mirroring Rails [`dot.rb:217-219`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L217-L219)
- plain object → `visitHash`
- everything else → `visitString` as a leaf (no crash)

The Node branch is checked before the AM-Attribute branch because Trails' `BindParam` exposes `valueBeforeTypeCast` as a method (via NodeExpression's duck-typing protocol); checking AM-shape first would steal the BindParam dispatch.

**Comment 2 — OptimizerHints lost its hints.** Real bug. `OptimizerHints` extends `Unary`, so dispatch fell back to `visitArelNodesUnary` which visits `expr` — but Trails stores the hint array on `hints` (`expr` stays `null`, per the unary.ts comment from #1014 review). Added a dedicated `visitArelNodesOptimizerHints(o)` that visitEdges 'hints', registered the node on the dispatch table.

Tests added pinning both fixes:
- BindParam wrapping a fake AM-Attribute renders `BindParam → fake → 42` via the `valueBeforeTypeCast` edge (no crash).
- OptimizerHints emits a `hints` edge with both hint values present in the graph.

1126/1126 arel tests passing; coverage holds at 820/820 (100%).

deanmarano:
Review #4 addressed (commit 2b50feb82) — but with a Rails-fidelity correction.

The reviewer suggested rendering `null`/`undefined` as `"nil"`. That's Ruby's [`nil.inspect`](https://ruby-doc.org/core/NilClass.html#method-i-inspect), but Rails' [`dot.rb`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L284-L291) doesn't use `inspect` — it uses `quote(field)` which is:

```ruby
def quote(string)
  string.to_s.gsub('"', '\\"')
end
```

`nil.to_s` is the **empty string** (`""`), not `"nil"`. So a Rails-faithful TS port should render `""` — that's what fixes the original bug (where `String(null)` was leaking JavaScript's `"null"` literal) without diverging from Rails.

Implemented as `value == null ? "" : String(value)` in `visitString`. Test pins the regex shape — `<f0>NilClass|<f1>` followed immediately by the closing `"`, with no characters in between, and explicitly asserts that neither `"null"` nor `"undefined"` appears in the output.

1127/1127 arel tests passing; coverage holds at 820/820 (100%).

deanmarano:
Review #5 addressed (commit b31db17ba) — both real bugs from my generic Rails port not matching Trails' actual node shapes:

**Comment 1 — Extract.** Trails' `Extract` extends `Unary` with `expr` + `field`; my `visitArelNodesExtract` was walking Rails' `expressions` + `alias`, emitting nil edges and never descending into the expression. Fixed to walk the Trails fields. (Rails' [`dot.rb:107-110`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L107-L110) walks `expressions` and `alias` because Rails' Extract has those fields — Trails' shape diverges from Rails here.)

**Comment 2 — Exists.** Trails' `Exists` is a standalone Node with `expressions: Node` + `alias` — not a Function subclass. Routing it to `visitArelNodesFunction` emitted a spurious `distinct` edge. Added a dedicated `visitArelNodesExists(o)` that walks only `expressions` and `alias`; updated the dispatch table.

Two tests added with positive + negative assertions: Extract emits `expr` and `field` edges and no `expressions`/`alias` ghost edges; Exists emits `expressions` and `alias` and no `distinct` ghost edge.

1127/1127 arel tests passing; coverage holds at 820/820 (100%).

deanmarano:
Review #6 addressed (commit 475b9cfed) — real bug.

Rails [`dot.rb#visit`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L237-L246) keys `@seen` by `o.object_id`. In Ruby that preserves per-instance identity for heap objects (two `String.new("foo")` get distinct entries) but dedupes language singletons (`nil` / `true` / `false` / Symbols / small Integers).

JS Map compares primitive keys by value, not by reference, so `seen.set("users", node)` made every later `"users"` string visit hit the cache and emit only an edge-redirect. Two distinct `Table("users")` instances rendered with a single shared `String|users` node — silently collapsing the graph.

Fixed by only memoizing reference-typed values (`typeof object === "object" && object !== null`). Primitives skip the seen-map entirely and get a fresh Dot node per visit. Reference-typed nodes still dedupe correctly via JS reference equality.

Test pins the regression: two `Table("users")` instances now produce two Table nodes AND two String `users` nodes. 1130/1130 arel tests passing; coverage holds at 820/820 (100%).

deanmarano:
Review #7 addressed (commit 6efa81b35) — all three:

**Comment 1 — null/undefined produced distinct NilClass nodes.** Right; Rails' `@seen` keys by `object_id`, and `nil.object_id` is a singleton, so multiple nil-valued fields collapse onto one NilClass node. My previous fix skipped primitive memoization entirely, which made each null hit produce a fresh node. Added a `NIL_SENTINEL` Symbol that null and undefined both map onto so they share one Dot node. Reference-typed values still memoize by identity; other primitives (strings, numbers, booleans) still skip memoization to avoid the false-collapse bug from review #6.

**Comment 2 — toDot edge filter was stale.** Right; with `visit()` always opening a Node and routing through `withNode` (which always sets `e.to`), the truthy-guard on edge emission was unreachable code. Replaced with an explicit assertion that throws `Dot: edge "X" has no destination node` if the invariant breaks — regressions surface loudly instead of silently dropping edges. Updated the `DotEdge.to` JSDoc to match the actual lifecycle.

**Comment 3 — no unit coverage for `snakeToCamel`.** Added `scripts/api-compare/conventions.test.ts` (10 tests) pinning standard snake-case, leading-underscore preservation, capital-segment camelCasing (`visit_Arel_Nodes_X`), run-of-underscore collapsing (`visit__regexp` / `__no_edges` / `__children`), and `rubyMethodToTs` special cases (`to_s`, `to_sql`, predicate `?`, bang `!`, setter `=`). A regression in any of those transforms now fails the dedicated test instead of silently breaking api:compare across hundreds of methods.

1130/1130 arel + 10/10 conventions tests passing; coverage holds at 820/820 (100%).

deanmarano:
Review #8 addressed (commit abe764bf9): right call. Rails' [`Dot#visit_edge`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/arel/visitors/dot.rb#L221-L223) does `o.send(method)` which raises `NoMethodError` on a typo. The TS port silently treated missing properties as `undefined`, emitting a NilClass leaf and hiding the visitor bug.

Added an `(method in o)` check that throws `undefined method 'X' for ClassName` when the field doesn't exist. `null` / `undefined` are still allowed when the property *is* declared but unset (legitimate case for unset `limit`/`offset` etc.).

Test added pinning the throw. All 1131 arel tests still pass — none of the 25+ existing visit_* methods reference a typo'd field, which the new check now proves.

Coverage holds at 820/820 (100%).

deanmarano:
Review #9 hit max review cycles — pausing further iteration. Summarized the three remaining nits in the PR description's new "Open follow-up nits" section so they can be picked up in a cleanup PR. None are correctness issues:

1. `as Node` casts in `visitArray` / `visitHash` / `visitChildren` are stale now that `Dot.visit` accepts `unknown` and routes by JS type.
2. `object instanceof Table` branch in `visit()` is redundant with `instanceof Node` (Table extends Node).
3. Same redundancy in `isPlainObject`'s `o instanceof Table` guard.

All three are static-typing/control-flow tidy-ups; behavior is equivalent.

deanmarano:
Review #9 nits cleaned up after all (commit be3771c54): dropped the `as Node` casts in visitArray/Children/Hash/Edge and the redundant `instanceof Table` checks in `visit()` and `isPlainObject`. PR description's follow-up section removed. 1131/1131 arel tests still passing; 820/820 (100%) coverage held.

deanmarano:
Review #10 hit max review cycles — pausing iteration. Summarized the five remaining nits in the PR description's new "Open follow-up nits" section:

1. `visitString` of a Symbol — should use `Symbol.keyFor` / `description` instead of `String(sym)` to match Ruby's `:foo.to_s` shape.
2. `@seen` doesn't memoize booleans/numbers/bigints/symbols — Rails dedupes singletons; would need a typed-prefix key strategy.
3. `visitArelNodesUpdateStatement` doesn't walk `groups` / `havings` (real gap — Trails' UpdateStatement has those fields).
4. `visitArelNodesDeleteStatement` same.
5. `classNameOf` JSDoc is stale (claims "String/Number/Boolean", emits Rails-style names).

3 and 4 are real correctness gaps for UPDATE/DELETE dot output with GROUP BY / HAVING clauses; the rest are fidelity polish. None affect the 820/820 coverage.

deanmarano:
Review #10 nits 2-5 addressed (commit ebbe34fa7); nit 1 (Symbol `String(sym)` → `description`) deferred per direction:

**3 & 4 — UPDATE/DELETE walk `groups` / `havings`.** Real correctness gaps. Added the `visitEdge` calls; tests pin the new edges in dot output for an UPDATE and a DELETE with GROUP BY / HAVING.

**2 — singleton dedupe for booleans / numbers / bigints / symbols.** Typed-prefix keys (`'boolean:true'`, `'number:42'`, `'bigint:N'`, plus Symbol identity) mirror Ruby's per-singleton `object_id`. Strings stay un-deduped (Ruby strings get distinct object_ids per instance — the case from review #6). Test pins: `visit(true)` twice → one TrueClass node; `visit(42)` twice → one Integer 42 node.

**5 — `classNameOf` JSDoc.** Updated to enumerate the actual emitted names (String / Integer / Float / TrueClass / FalseClass / NilClass / Symbol / Time).

Follow-up section in the PR body removed. 1134/1134 arel tests passing; 820/820 (100%) coverage held.

deanmarano:
Review #11 hit max review cycles — pausing iteration. Summarized the two remaining nits in the PR description's new "Open follow-up nits" section:

1. `visitString` of a Symbol — still emits `String(sym)` ('Symbol(foo)') instead of Ruby's `Symbol#to_s` ('foo'). Carried over from review #10 (deferred per user direction).
2. JS plain objects routed through `visitHash` are labeled `Object` (JS ctor name) rather than Rails' `Hash`. `classNameOf` should special-case plain objects to return 'Hash'.

Both are output-shape polish; neither affects the 820/820 coverage.

PR #1015, opened 2026-04-29
feat(activemodel): port Numericality coercion privates (parseFloat, round, isIsNumber, isIsInteger, isIsHexadecimalLiteral, optionAsNumber)

deanmarano:
## Summary
Track A4a of [docs/activemodel-privates-100-plan.md](../blob/main/docs/activemodel-privates-100-plan.md). Mirrors `ActiveModel::Validations::NumericalityValidator` (`numericality.rb`) coercion pipeline:

- `INTEGER_REGEX` (`/^[+-]?\d+(?![\s\S])/` — true `\z`-equivalent, rejects `"1\n"`) + `HEXADECIMAL_REGEX` constants — `numericality.rb:18-20`.
- `parseFloat(num, precision, scale)` — `numericality.rb:86-88`. `round(num, scale)` then `toPrecision`.
- `round(num, scale)` — delegates to activesupport `RoundingHelper` (half-away-from-zero, single source of truth in the repo).
- `isIsNumber(rawValue, precision, scale)` — `numericality.rb:94-100`. Honors `only_numeric`; restricts to `string|number` (Rails `Kernel.Float` raises `TypeError` on Date/boolean/objects); rejects blank, leading-whitespace hex, and `0b…`/`0o…` literals.
- `isIsInteger(rawValue)` — `numericality.rb:102-104`.
- `isIsHexadecimalLiteral(rawValue)` — `numericality.rb:106-108`. Anchored, no leading whitespace.
- `optionAsNumber(record, optionValue, precision, scale)` — `numericality.rb:67-69`. **Closes the `resolveValue` consumption gap** flagged for Numericality in PR #971. Anchored hex returns `undefined` (Rails skip semantic); `0b/0o` and trim-only-hex throw the consistent `Resolved numericality option must be numeric: …` error.

Privates attached to `NumericalityValidator.prototype` (not class fields) so they survive `EachValidator`'s super-time `checkValidity()` call — same bootstrapping pattern as PRs #994 / #1002 / #1009. `@internal` JSDoc tags satisfy `blazetrails/rails-private-jsdoc`. Names follow `extract-ts-api` conventions: `is_number?` → `isIsNumber`, etc. (the doubled `is` prefix is fixed in a follow-up PR — see Outstanding section below).

## Out of scope (Track A4b)
`filtered_options`, `allow_only_integer?`, `prepare_value_for_validation`, `record_attribute_changed_in_place?` — the dispatch-side helpers. Saved for the next PR.

## Impact
[...]

PR #1028, opened 2026-04-30
feat(activemodel): port validator-bundle privates (Confirmation, Acceptance, Callbacks)

deanmarano:
## Summary
Track A5 of [docs/activemodel-privates-100-plan.md](../blob/main/docs/activemodel-privates-100-plan.md) — final validator-cluster track. Closes 6 privates across 3 small files; each row goes 2/4 → **4/4 (100%)**.

### Confirmation (`confirmation.rb` 2/4 → 4/4)
- **`isConfirmationValueEqual(record, attribute, value, confirmed)`** — `confirmation.rb:32-38`. Extracts the inline equality check (case-sensitive vs `casecmp`) and is dispatched through `this.isConfirmationValueEqual` from `validateEach`.
- **`setupBang(klass)`** — `confirmation.rb:21-29`. Defines virtual `${attribute}Confirmation` reader/writer accessors on the host class prototype with a per-instance `_${attr}Confirmation` backing slot; Rails uses `attr_reader`/`attr_writer` the same way.

### Acceptance (`acceptance.rb` 2/4 → 4/4)
- **`isAcceptableOption(value)`** — `acceptance.rb:24-26`. Extracts the `Array(options[:accept]).include?(value)` normalization (with the `hasOwn` check on `accept` that PR #976 established for explicit-`nil` semantics) and is dispatched through `this.isAcceptableOption`.
- **`setupBang(klass)`** — `acceptance.rb:18-22`. Eagerly installs accessors (no `method_missing` in TS), per-instance backing slot, skips attributes the host already defines.

### Callbacks (`callbacks.rb` 2/4 → 4/4)
- **`setOptionsForCallback(options)`** — `callbacks.rb:99-110`. Normalizes `on:` to an array and prepends the validation-context guard to `if:`. In-place mutation matches Rails.
- **`runValidationsBang()`** — `callbacks.rb:113-115`. Wraps a `runValidations()` block in `_runValidationCallbacks`. Surfaces in the `Callbacks` contract; trails Model already does this internally — this gives downstream hosts a typed reference.

All host-class privates attached to the validator prototypes (with `@internal` JSDoc tags) using the super-time bootstrapping pattern from PRs #994 / #1002 / #1009 / #1015 / #1026.

## Impact
- `pnpm api:compare --privates --package activemodel`: confirmation.rb / acceptance.rb / callbacks.rb each 2/4 → **4/4 (100%)** (+6). Overall 499/625 → **505/625 (80.8%)**.
- `pnpm api:compare --package activemodel`: stays **448/448 (100%)**.
- `pnpm test:compare --package activemodel`: stays 959/963.

## Track A status
🎉 **Track A complete.** Validator family (Clusivity / Inclusion / Exclusion / Format / Length / Numericality / Confirmation / Acceptance / Callbacks) all at 100% privates. The next bucket (Track B — privates beyond validators: serialization, type primitives, model.rb / validations.rb) is the next push if you want to keep going.

## Test plan
- [x] `pnpm vitest run packages/activemodel` — 1604 / 1604 passing
- [x] `pnpm api:compare --package activemodel` — 448/448
- [x] `pnpm api:compare --privates --package activemodel` — 505/625 (+6)
- [x] `pnpm test:compare --package activemodel` — 959/963 (0)

### DIFF
--- a/packages/activemodel/src/validations/_accessor.ts
+++ b/packages/activemodel/src/validations/_accessor.ts
@@ -1,0 +1,62 @@
+/**
+ * Shared internal helper for the validator setupBang ports
+ * (acceptance.ts, confirmation.ts). TS-only plumbing — no Rails
+ * counterpart. Walks the prototype chain to capture the first-found
+ * accessor shape for `name` so callers can install only the missing
+ * half of a getter/setter pair without shadowing inherited accessors.
+ */
+
+export interface InheritedAccessor {
+  hasGetter: boolean;
+  hasSetter: boolean;
+  getter?: (this: object) => unknown;
+  setter?: (this: object, value: unknown) => void;
+}
+
+/**
+ * Distinguishes accessor descriptors (`get`/`set`) from data
+ * descriptors (`"value" in desc` — handles the `value: undefined`
+ * case correctly). For data properties, synthesizes a getter that
+ * reads through the captured prototype via `Reflect.get` (avoids
+ * recursing into the accessor the caller is about to install), and a
+ * setter that writes via own-property `defineProperty`.
+ */
+export function inspectAccessor(prototype: object, name: string): InheritedAccessor {
+  let proto: object | null = prototype;
+  while (proto && proto !== Object.prototype) {
+    const desc = Object.getOwnPropertyDescriptor(proto, name);
+    if (desc) {
+      if ("value" in desc || "writable" in desc) {
+        const inheritedProto = proto;
+        const enumerable = desc.enumerable ?? true;
+        const configurable = desc.configurable ?? true;
+        return {
+          hasGetter: true,
+          hasSetter: desc.writable !== false,
+          getter() {
+            return Reflect.get(inheritedProto, name, this);
+          },
+          setter:
+            desc.writable !== false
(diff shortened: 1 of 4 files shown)


deanmarano:
Re: comment #4 (`setOptionsForCallback` leaving `on` on options)

Keeping `on` to mirror Rails. From `activemodel/lib/active_model/validations/callbacks.rb:99-110`:

```ruby
def set_options_for_callback(options)
  if options.key?(:on)
    options[:on] = Array(options[:on])
    options[:if] = [
      ->(o) { options[:on].intersect?(Array(o.validation_context)) },
      *options[:if]
    ]
  end
end
```

Rails normalizes `options[:on]` and leaves it on the hash; `set_callback` ignores unknown keys. The lambda Rails builds closes over `options[:on]` (re-reads on every call), so deleting it would break the guard there. This is an `@internal` Rails-private port — host code that wants to feed the result into trails' `CallbackChain.register` (which is stricter than AS::Callbacks) should strip `on` itself, not have the helper diverge from Rails. (My port's lambda captures `onArr` by value, so a deletion wouldn't break ours, but I'd rather keep behavior identical to Rails than save callers one `delete`.)

PR #1034, opened 2026-04-30
fix(arel): SQLite strips Grouping operands from set operations

deanmarano:
## Summary

SQLite rejects parens around SELECT operands of UNION/INTERSECT/EXCEPT. This mirrors Rails' `Arel::Visitors::SQLite#infix_value_with_paren` (sqlite.rb:39-60) by overriding the four set-op visitors in the SQLite visitor to unwrap a `Grouping` wrapper from each operand before visiting.

Relaxes the four base set-op visitors in `to-sql.ts` from `private` to `protected` so the SQLite subclass can override them.

## Test plan

- [x] `pnpm vitest run packages/arel` — 1193 passing
- [x] New tests cover Union/UnionAll/Intersect/Except with and without Grouping wrappers

### DIFF
--- a/packages/arel/src/visitors/sqlite.test.ts
+++ b/packages/arel/src/visitors/sqlite.test.ts
@@ -65,4 +65,49 @@ describe("SqliteTest", () => {
       expect(sql).toContain('"posts"."user_id"');
     });
   });
+
+  describe("set operations", () => {
+    // SQLite rejects parens around SELECT operands of UNION/INTERSECT/EXCEPT.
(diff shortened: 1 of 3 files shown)


deanmarano:
Both comments suggest recursive multi-layer unwrap, but Rails `Arel::Visitors::SQLite#infix_value_with_paren` (sqlite.rb:42, 51) does a single-layer unwrap: `o.left.is_a?(Nodes::Grouping) ? o.left.expr : o.left`. No recursion. Keeping single-layer unwrap to mirror Rails — `Grouping(Grouping(...))` is not produced by the SelectManager API in either Rails or trails.

PR #1035, opened 2026-04-30
fix(arel): wire set-op visitors through infixValueWithParen + flatten

deanmarano:
## Summary

Base `ToSql` Union/UnionAll/Intersect/Except visitors now delegate to the existing `infixValueWithParen` helper, mirroring `to_sql.rb#visit_Arel_Nodes_Union` and friends. Output shape becomes the Rails-faithful `( a UNION b UNION c )` — flat, single outer pair of parens, inner SELECT operands visited via `groupingParentheses(false)`.

The SQLite visitor now overrides `infixValueWithParen` itself (mirroring `sqlite.rb#infix_value_with_paren`) instead of the four set-op visitors. Same end behavior as PR #1034, but matches Rails' inheritance shape and inherits flatten-by-class for free.

## Test plan

- [x] `pnpm vitest run packages/arel` — 1196 passing
- [x] Spot-checked AR relations + relation/with tests — green
- [x] New strict-shape tests for nested UNION/UNION ALL/INTERSECT/EXCEPT in `to-sql.test.ts`
- [x] SQLite test for nested unions

### DIFF
--- a/packages/arel/src/visitors/sqlite.test.ts
+++ b/packages/arel/src/visitors/sqlite.test.ts
@@ -77,31 +77,40 @@ describe("SqliteTest", () => {
       const sql = new Visitors.SQLite().compile(node);
       expect(sql).not.toContain("((");
       expect(sql).toContain("UNION");
-      expect(sql).toBe('(SELECT * FROM "users" UNION SELECT * FROM "users")');
+      expect(sql).toBe('( SELECT * FROM "users" UNION SELECT * FROM "users" )');
     });
 
     it("UNION ALL strips Grouping wrapped operands", () => {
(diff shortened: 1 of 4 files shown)


deanmarano:
Comment 1 (ORDER/LIMIT/OFFSET coverage): added 3 regression tests covering wrapped SELECT operands per Rails `grouping_parentheses(..., false)` + `require_parentheses?`.

Comment 2 (recursive unwrap): declining — Rails `Arel::Visitors::SQLite#infix_value_with_paren` (sqlite.rb:42, 51) does single-layer unwrap (`o.left.is_a?(Nodes::Grouping) ? o.left.expr : o.left`), no recursion. `Grouping(Grouping(...))` is not produced by the SelectManager API; matching Rails fidelity over hypothetical defense.

PR #1040, opened 2026-04-30
feat(activesupport): time-ext period bounds return Temporal.Instant

deanmarano:
## Summary

F-6a of the Temporal migration follow-ups: 15 period-bound helpers and 5 range delegators in `time-ext.ts` now return `Temporal.Instant`.

- `beginningOfDay`, `middleOfDay`, `endOfDay`
- `beginningOfHour`, `endOfHour`
- `beginningOfMinute`, `endOfMinute`
- `beginningOfWeek`, `endOfWeek`
- `beginningOfMonth`, `endOfMonth`
- `beginningOfQuarter`, `endOfQuarter`
- `beginningOfYear`, `endOfYear`
- `allDay`, `allWeek`, `allMonth`, `allQuarter`, `allYear`

Inputs remain `Date` for now — F-6b/c/d will flip the rest. Internal arithmetic still routes through cloned `Date` mutators to preserve exact local-tz semantics; the result is wrapped via `Temporal.Instant.fromEpochMilliseconds(d.getTime())`.

`nextWeek` / `prevWeek` (which used `beginningOfWeek` for Date arithmetic) now use a private `_beginningOfWeekDate` helper so they keep returning `Date` until F-6b.

## Test plan

- [x] `pnpm vitest run packages/activesupport` — 3639/3639 passing
- [x] `pnpm build` clean
- [x] `pnpm lint` clean
- [x] Diff: 196 +/146 -, 5 files (slightly over the 300 LOC target — scope is the minimum coherent unit since the 20 helpers share test-file infrastructure)

### DIFF
--- a/packages/activesupport/src/core-ext/date-ext.test.ts
+++ b/packages/activesupport/src/core-ext/date-ext.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect } from "vitest";
+import { Temporal } from "../temporal.js";
 import {
   beginningOfDay,
   middleOfDay,
@@ -30,6 +31,10 @@ import {
 
 function d(year: number, month: number, day: number, hour = 0, min = 0, sec = 0, ms = 0): Date {
   return new Date(year, month - 1, day, hour, min, sec, ms);
+}
+
+function asDate(instant: Temporal.Instant): Date {
+  return new Date(instant.epochMilliseconds);
 }
 
 describe("DateExtBehaviorTest", () => {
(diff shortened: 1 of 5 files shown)


deanmarano:
CI failure on SQLite Tests is **unrelated to this PR's diff**.

The failures are in `packages/activerecord/src/token-for.test.ts`:
- `does not find record when token has expired`
- `does not find record when expires_in is different`

Both fail with `RangeError: unsupported fractional value 0.001` — these tests pass `expiresIn: 0.001` (fractional seconds), but `MessageVerifier#generate` (changed in F-4 / #1037) now does `Temporal.Now.instant().add({ seconds: options.expiresIn })`, and `Temporal.Instant#add` rejects non-integer time-unit values.

This PR (F-6a) only touches `time-ext.ts` period-bound helpers — none of those are reachable from `token-for.test.ts`. The regression is in F-4's territory (`message-verifier.ts:65`), where `add({ seconds })` should be `add({ milliseconds: ms * 1000 })` or `add({ nanoseconds: BigInt(Math.round(seconds * 1e9)) })` to accept fractional seconds.

Will track as an F-4 follow-up. Not blocking this PR.

PR #1045, opened 2026-04-30
fix(arel): compileUpdate/compileDelete always assign key

deanmarano:
## Summary

- Drops `if (key !== null)` guards in `SelectManager#compileUpdate` and `#compileDelete` so `key` (including `null`) is always assigned to the resulting `UpdateManager`/`DeleteManager`.
- Matches Rails `Arel::Crud` ([crud.rb](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/crud.rb)) where `um.key = key` and `dm.key = key` are unconditional.

Refs [`docs/arel-alignment-plan.md` PR 17](../blob/main/docs/arel-alignment-plan.md).

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1215/1215 passing, including new `compileUpdate(values, null)` / `compileDelete(null)` assertions verifying `manager.key === null`.

### DIFF
--- a/packages/arel/src/crud.test.ts
+++ b/packages/arel/src/crud.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { Table, InsertManager, UpdateManager, DeleteManager } from "./index.js";
+import { Table, InsertManager, UpdateManager, DeleteManager, SelectManager } from "./index.js";
 
 describe("crud", () => {
   const users = new Table("users");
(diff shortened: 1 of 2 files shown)


deanmarano:
Addressed both review comments in 269dd6a78:

1. The test now spies on the prototype's `key` setter via `Object.defineProperty`, so a null key being skipped is observable. Verified by temporarily reverting the impl: tests fail as expected.
2. Updated the comment to describe the change as Rails parity / explicit assignment rather than "clearing a prior value".

deanmarano:
Switched to `vi.spyOn(UpdateManager.prototype, "key", "set")` + `afterEach(vi.restoreAllMocks)` per the suggestion. Spying on `UpdateManager.prototype` / `DeleteManager.prototype` directly (rather than `TreeManager.prototype`) because TS `declare key` in each subclass installs the accessor descriptor on the subclass prototype. Verified regression detection by temporarily restoring the guard.

PR #1048, opened 2026-04-30
fix(arel): wave-1 fidelity bundle (from/key=/lower/cast/UnaryOperation)

deanmarano:
## Summary

Bundles four small Rails-fidelity fixes from [`docs/arel-alignment-plan.md`](../blob/main/docs/arel-alignment-plan.md), all independent and theme-coherent (touch one method each on disjoint files).

- **PR 13** — `SelectManager#from` now routes a `Join` to `core.source.right` rather than overwriting `core.source.left`, matching [`select_manager.rb`'s `case … when Nodes::Join`](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/select_manager.rb).
- **PR 18** — `TreeManager#key=` calls `Nodes.build_quoted` on scalars and maps over arrays, matching `tree_manager.rb`.
- **PR 19** — `FactoryMethods#lower` wraps non-Node arguments via `build_quoted`; `#cast` delegates to `name.as(type)` so the alias is a retryable `SqlLiteral`, matching `factory_methods.rb`.
- **PR 20** — `visit_Arel_Nodes_UnaryOperation` no longer `trim()`s the operator, matching `visitors/to_sql.rb`'s verbatim `" #{operator} "` emission.

99 insertions / 23 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1222/1222 (added one targeted test per PR).
- [x] `pnpm -w build` — clean.

### DIFF
--- a/packages/arel/src/factory-methods.test.ts
+++ b/packages/arel/src/factory-methods.test.ts
@@ -42,6 +42,15 @@ describe("TestFactoryMethods", () => {
     expect(fn.name).toBe("LOWER");
   });
 
+  // Mirrors Rails: `lower(column)` wraps non-Node arguments via
+  // `Nodes.build_quoted` (factory_methods.rb), so a string column resolves
+  // to LOWER('name') rather than rendering "[object Object]".
+  it("lower wraps non-Node arguments via buildQuoted", () => {
+    const fn = users.lower("name");
+    expect(fn.expressions[0]).toBeInstanceOf(Nodes.Quoted);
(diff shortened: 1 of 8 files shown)


deanmarano:
The CI failure on run 25175246578 was concurrency cancellation, not a real test failure: PostgreSQL/MariaDB jobs were cancelled with "Canceling since a higher priority waiting request for ci-CI-refs/pull/1048/merge exists" because a newer push superseded that run. The successor run (25175543067) is now in progress.

PR #1050, opened 2026-04-30
fix(arel): Statement ctors take relation; extract wraps in array

deanmarano:
## Summary

Bundles two coherent Rails-fidelity fixes (both reshape node ctors). Both are independent and small.

- **PR 21** — Statement ctors take a `relation` arg.
  - `SelectStatement(relation)` / `SelectCore(relation)` mirror [`select_statement.rb`](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/nodes/select_statement.rb) and [`select_core.rb`](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/nodes/select_core.rb): the seed `SelectCore` is constructed pre-bound to a FROM target.
  - `InsertStatement(relation)` mirrors `insert_statement.rb`.
  - `DeleteStatement(relation, wheres = [])` adds the second arg per `delete_statement.rb`.
  - `UpdateStatement` already matched Rails.
- **PR 22** — `Expressions#extract` / `Attribute#extract` now construct `new Extract([self], field)` (single-element array), matching `expressions.rb`'s `Nodes::Extract.new [self], field`. The `visit_Arel_Nodes_Extract` visitor renders an array operand via `inject_join`, so the single-element form produces the same SQL as the prior bare-expression form.

55 insertions / 14 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1226/1226.
- [x] `pnpm -w build` clean.

### DIFF
--- a/packages/arel/src/attributes/attribute.ts
+++ b/packages/arel/src/attributes/attribute.ts
@@ -519,7 +519,8 @@ export class Attribute extends Node {
   // -- Extract --
 
   extract(field: string): Extract {
-    return new Extract(this, field);
+    // Mirrors Rails: `Nodes::Extract.new [self], field` (expressions.rb).
+    return new Extract([this], field);
   }
 
   // -- Null handling --
--- a/packages/arel/src/expressions.ts
+++ b/packages/arel/src/expressions.ts
@@ -35,6 +35,7 @@ export const Expressions: ExpressionsModule = {
     return new Avg([this]);
   },
   extract(this: Node, field: string): Extract {
-    return new Extract(this, field);
+    // Mirrors Rails: `Nodes::Extract.new [self], field` (expressions.rb).
+    return new Extract([this], field);
   },
 };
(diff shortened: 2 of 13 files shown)


deanmarano:
Addressed both comments in b147ec8e0: widened `Unary.expr` to include `Node[]` so `Extract` no longer needs `as unknown as Node` and the visitor calls `visitArray(node.expr)` directly. No behavioral change; 1226 arel tests still pass and build is clean.

deanmarano:
Both addressed in e97202db7:

1. `visitNodeOrValue` now routes `Array.isArray(v)` through `visitArray` before the primitive-stringification branches, so any Unary visitor that uses `visitNodeOrValue` for its expr now renders arrays via Rails' `visit_Array` semantics instead of `"[object Object]"`.
2. Reworded the `OptimizerHints` doc — the dedicated `hints` field exists for element typing (`string | SqlLiteral` rather than `Node`), not because `expr` couldn't be an array.

PR #1055, opened 2026-04-30
fix(arel): SelectManager#distinct(value), lateral order, comment array form

deanmarano:
## Summary

Three small Rails-fidelity tweaks (PR 15 in [`docs/arel-alignment-plan.md`](../blob/main/docs/arel-alignment-plan.md)).

- **distinct** — `SelectManager#distinct(value = true)` now clears the set quantifier when `value` is falsy (matches [`select_manager.rb`](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/select_manager.rb)).
- **lateral** — `SelectManager#lateral(name?)` now produces `Lateral.new(as(name))` so the `TableAlias` lives **inside** the `Lateral`, not vice versa (was reversed). The base `to-sql` `Lateral` visitor now wraps in parens only when the operand is a `SelectStatement`, matching Rails' `grouping_parentheses`. The PostgreSQL override is now redundant and was removed.
- **comment** — `Nodes::Comment.new(values)` now takes a single `string[]` arg; `SelectManager#comment(*values)` passes the values array through, matching Rails.

The lateral change is the only SQL-output-affecting fix:
- `lateral("t")` now renders `LATERAL (SELECT …) t` (Rails-faithful) instead of the prior `LATERAL (SELECT …) "t"`. Updated the postgres visitor test accordingly; deleted two obsolete dialect-divergence audit-followup tests since the base now matches Rails directly.

76 insertions / 45 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1228/1228 (5 new tests under `describe("lateral")` and `it("distinct(false) clears the set quantifier")` etc.).
- [x] AR's `from(lateral_subquery)` flow exercised via existing arel/AR suites; no fixture changes needed.

### DIFF
--- a/packages/arel/src/nodes/comment.ts
+++ b/packages/arel/src/nodes/comment.ts
@@ -8,7 +8,7 @@ import { Node, NodeVisitor } from "./node.js";
 export class Comment extends Node {
   readonly values: string[];
 
-  constructor(...values: string[]) {
+  constructor(values: string[]) {
     super();
     this.values = values;
   }
--- a/packages/arel/src/select-manager.test.ts
+++ b/packages/arel/src/select-manager.test.ts
@@ -829,6 +829,46 @@ describe("SelectManagerTest", () => {
     );
   });
 
+  // Mirrors Rails: `distinct(value=true)` clears the set quantifier when
+  // value is falsy (select_manager.rb).
+  it("distinct(false) clears the set quantifier", () => {
+    const mgr = users.project(users.get("name")).distinct();
+    expect(mgr.toSql()).toContain("DISTINCT");
+    mgr.distinct(false);
+    expect(mgr.toSql()).not.toContain("DISTINCT");
+  });
+
+  describe("lateral", () => {
+    // Mirrors Rails: `lateral` returns `Lateral.new(ast)` when no name is
+    // given (select_manager.rb), so the visitor renders `LATERAL (...)`.
+    it("returns a Lateral wrapping the SELECT", () => {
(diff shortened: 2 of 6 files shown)


deanmarano:
Both addressed in 10df8419e:

1. `distinct` now narrows the param to `boolean | null` and clears only when `value === false || value == null`, matching Ruby's `if value` (where `0` / `""` are truthy).
2. Reworded the Lateral visitor comment — the `TableAlias` built by `SelectManager#as` renders its alias *bare* (per `visitArelNodesTableAlias`'s Grouping branch), not quoted; the prior wording implied quoted output.

deanmarano:
Both addressed in 299a7b71b:

1. `visitArelNodesLateral` now delegates to the existing `groupingParentheses` helper — same Rails-faithful logic, no duplication, picks up any future drift in `require_parentheses?` automatically.
2. `distinct` check tightened to `value === false || value === null` and the docstring trimmed — the `= true` default means `distinct(undefined)` would resolve to `distinct(true)` and never reach the falsy branch, so listing `undefined` was misleading.

deanmarano:
Both addressed in 0aed22728:

1. `distinct` param widened back to `unknown` (the prior `boolean | null` was inconsistent with the doc & runtime intent — it blocked Rails-valid truthy values like `0` or `""`). Runtime check stays `value === false || value == null` so only those two clear DISTINCT.
2. Test comment now spells out "`false`/`nil` only" instead of "falsy" to disambiguate from JS truthiness.

PR #1057, opened 2026-04-30
fix(arel): take/skip raw amount + null clear; optimizerHints AST node

deanmarano:
## Summary

Bundles two coherent SelectManager AST fixes from [`docs/arel-alignment-plan.md`](../blob/main/docs/arel-alignment-plan.md) (PRs 6 + 14). Both write-side; the read-side `limit`/`offset` getter change is the still-deferred PR 7 (wave 3).

- **PR 6** — `take(amount)` / `skip(amount)` now mirror [`select_manager.rb`](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/select_manager.rb): `null` clears `ast.limit`/`offset`; raw amounts flow into `Nodes::Limit` / `Nodes::Offset` directly without a `Quoted` wrap. Adds `limit=` / `offset=` setters per Rails' `alias :limit= :take` / `alias :offset= :skip`.
  - Consumer-visible: `(mgr.ast.limit as Limit).expr` is now a primitive (was `Quoted`). The `mgr.limit` getter still returns the `Limit` node — the breaking getter return-type change is PR 7.
- **PR 14** — `optimizerHints(...hints)` now wraps the values in a `Nodes::OptimizerHints` node and stores it on `SelectCore#optimizerHints` (was a bare `string[]`). `emitOptimizerHints` delegates to the dedicated visitor; sanitization (newlines, comment delimiters, `--` line comments) and empty-hint filtering preserved.

123 insertions / 28 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1236/1236 (8 new tests under `take`/`skip`/`limit= / offset= setters` + two assertions for the OptimizerHints node shape).
- [x] `pnpm exec vitest run packages/activerecord/` — 9947 passed, no regressions.

### DIFF
--- a/packages/arel/src/nodes/select-core.ts
+++ b/packages/arel/src/nodes/select-core.ts
@@ -1,5 +1,6 @@
 import { Node, NodeVisitor } from "./node.js";
 import { JoinSource } from "./join-source.js";
+import type { OptimizerHints } from "./unary.js";
 
 /**
  * SelectCore — the core of a SELECT statement (projections, from, where, etc.).
@@ -15,8 +16,9 @@ export class SelectCore extends Node {
   windows: Node[];
   setQuantifier: Node | null;
   // Mirrors Rails: `@ctx.optimizer_hints` is an `OptimizerHints` node (or
-  // nil) — not a bare array (select_core.rb).
-  optimizerHints: Node | null;
+  // nil) — not a bare array (select_core.rb). Narrowed to the concrete
+  // node type so the SQL visitor can rely on the `/*+ … */` formatting.
+  optimizerHints: OptimizerHints | null;
   comment: Node | null;
 
   constructor(relation: Node | null = null) {
--- a/packages/arel/src/nodes/select-core.ts
+++ b/packages/arel/src/nodes/select-core.ts
@@ -14,7 +14,9 @@ export class SelectCore extends Node {
   havings: Node[];
   windows: Node[];
   setQuantifier: Node | null;
-  optimizerHints: string[];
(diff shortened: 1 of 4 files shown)


deanmarano:
Both addressed in a45fff918:

1. `SelectCore.optimizerHints` narrowed to `OptimizerHints | null` (type-only import to avoid runtime cycles).
2. Stale comment on `visitArelNodesOptimizerHints` updated — it now says SelectCore stores an OptimizerHints node and `emitOptimizerHints` delegates here, rather than describing the prior bare-string[] inline storage.

deanmarano:
1. Fixed in dc8298b28 — `SqlLiteral` hints now pass through unchanged (matches `sanitizeAsSqlComment`'s contract used elsewhere); only plain strings run through `sanitizeHint`.

2. Not changing: `OptimizerHints.hints` is declared `readonly` as a `ReadonlyArray` ([nodes/unary.ts:90](../blob/main/packages/arel/src/nodes/unary.ts#L90)), so a clone-time copy wouldn't add real safety — the hints array can't legally mutate without a casts-bypass that already breaks the type contract. The node itself is also a plain immutable AST element. Cloning by reference matches how Trails clones other readonly leaf nodes (e.g. SelectStatement#clone copies `limit`/`offset`/etc. by reference).

deanmarano:
Both addressed in 702f7f52e:

1. `optimizerHints` signature widened to `(string | SqlLiteral)[]` — the `SqlLiteral` escape-hatch is now reachable from the fluent API without casting.
2. `collectOptimizerHints` doc comment updated to describe the new `OptimizerHints`-node storage on SelectCore.

PR #1061, opened 2026-04-30
fix(arel): InsertManager shape parity

deanmarano:
## Summary

PR 12 from [`docs/arel-alignment-plan.md`](../blob/main/docs/arel-alignment-plan.md) — bring `InsertManager` AST in line with Rails ([insert_manager.rb](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/insert_manager.rb)).

- `insert(fields)`:
  - early-returns on `null`/empty (Rails: `return if fields.empty?`)
  - string form stores `Nodes::SqlLiteral.new(fields)` on `ast.values` (was missing)
  - infers `ast.relation` from `fields[0][0].relation` when not yet set (Rails: `@ast.relation ||= fields.first.first.relation`)
  - drops the `Quoted` wrap on values; the visitor's `visitNodeOrValue` path renders raw primitives
- Replaces the prior `values(list)` method with a `set values(val)` setter, matching Rails' `attr_writer :values`. Atomic in-tree migration: AR's `schema-migration.ts` and the `insert-manager.test.ts` suite updated to `mgr.values = …`.
- `select(selectManager)` now stores the manager itself (matching Rails) rather than unwrapping to `.ast`. `InsertStatement#select` widens to `Node | { ast; toSql() } | null`; the visitor routes through `visitNodeOrValue` so the SelectManager-shaped duck-type lands in `visitArelSelectManager`.
- Insert visitor prefers `node.values` over `node.select` when both are present, matching Rails' `to_sql.rb` pattern.
- `ValuesList.rows` widened to `unknown[][]` to carry raw primitives; `visitArelNodesValuesList` uses `visitNodeOrValue` per row cell.

118 insertions / 45 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1241/1241 (5 new tests under `insert (Rails parity)` and `select (Rails parity)`).
- [x] `pnpm exec vitest run packages/activerecord/` — 9947 passed, no regressions (one site in `schema-migration.ts` updated to the setter form).
- [x] `pnpm -w build` clean.

### DIFF
--- a/packages/activerecord/src/schema-migration.ts
+++ b/packages/activerecord/src/schema-migration.ts
@@ -199,7 +199,7 @@ export class SchemaMigration {
       const im = new InsertManager(this.arelTable);
       const rows = toInsert.map((v) => [new Nodes.Quoted(v)]);
       im.ast.columns = [col];
-      im.values(im.createValuesList(rows));
+      im.values = im.createValuesList(rows);
       await this._adapter.executeMutation(im.toSql());
     }
   }
--- a/packages/arel/src/insert-manager.test.ts
+++ b/packages/arel/src/insert-manager.test.ts
@@ -9,12 +9,10 @@ describe("InsertManagerTest", () => {
       const mgr = new InsertManager();
       mgr.into(users);
       mgr.ast.columns = [users.get("name"), users.get("age")];
-      mgr.values(
-        new Nodes.ValuesList([
-          [new Nodes.Quoted("dean"), new Nodes.Quoted(30)],
-          [new Nodes.Quoted("sam"), new Nodes.Quoted(25)],
-        ]),
-      );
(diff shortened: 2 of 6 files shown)


deanmarano:
Not changing — Trails' `push()` matches Rails exactly. From [insert_manager.rb:31-32](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/insert_manager.rb):

```ruby
fields.each do |column, value|
  @ast.columns << column   # << is push, not assign
  values << value
end
```

Rails accumulates columns across calls (the `@ast.columns = []` initializer in InsertStatement is set once at construction). Switching to `= [col, ...]` would diverge from Rails. The Rails contract is "call `insert` once per insertion" — same as Trails.

PR #1062, opened 2026-04-30
fix(arel): UpdateManager#set wraps columns in UnqualifiedColumn

deanmarano:
## Summary

PR 8 from [`docs/arel-alignment-plan.md`](../blob/main/docs/arel-alignment-plan.md) — bring `UpdateManager#set` AST in line with Rails ([update_manager.rb](../blob/main/scripts/api-compare/.rails-source/activerecord/lib/arel/update_manager.rb)).

- `UpdateManager#set([[col, val], …])` now produces `Assignment(UnqualifiedColumn(col), value)` with the value passed raw — the visitor's `visitNodeOrValue` quotes primitives.
- Removes the stateful `_inUpdateSet` flag from the `to-sql` visitor. LHS table-qualifier stripping is now structural via `UnqualifiedColumn`. The Assignment visitor reduces to plain `visit(left) = visit(right)`, matching Rails verbatim.

### Side fix (caught by AR regression)

`visitNodeOrValue`'s unknown-object fallback now defers to `quote()` instead of concatenating `String(v)` raw. This matters for non-`Date` date-likes (e.g. `Temporal.Instant` from AR's timestamp helpers) that flow through `UPDATE … SET` assignments without a `Quoted` wrap — the previous fallback inlined them as bare ISO tokens (\`SET "updated_at" = 2026-04-30T19:30:00Z\`), tripping SQLite's `unrecognized token` error. Counter-cache + collection-cache tests now pass.

50 insertions / 18 deletions.

## Test plan

- [x] `pnpm exec vitest run packages/arel/` — 1244/1244 (3 new tests under `set`).
- [x] `pnpm exec vitest run packages/activerecord/` — 9947 passed, no regressions (counter-cache + collection-cache verified).
- [x] `pnpm -w build` clean.

### DIFF
--- a/packages/arel/src/update-manager.test.ts
+++ b/packages/arel/src/update-manager.test.ts
@@ -103,6 +103,35 @@ describe("UpdateManagerTest", () => {
       mgr.set(new Nodes.BoundSqlLiteral("name = ?", ["dean"], {}));
       expect(mgr.toSql()).toBe(`UPDATE "users" SET name = 'dean'`);
     });
+
+    // Mirrors Rails: `set` wraps each LHS in `Nodes::UnqualifiedColumn`
+    // (update_manager.rb), so the visitor strips the table qualifier
+    // structurally rather than via a stateful flag.
+    it("wraps each column in an UnqualifiedColumn", () => {
+      const mgr = new UpdateManager();
+      mgr.table(users);
+      mgr.set([[users.get("name"), "dean"]]);
+      const assignment = mgr.ast.values[0] as Nodes.Assignment;
+      expect(assignment).toBeInstanceOf(Nodes.Assignment);
+      expect(assignment.left).toBeInstanceOf(Nodes.UnqualifiedColumn);
+    });
+
+    // Mirrors Rails: values pass through raw (no Quoted wrap); the
+    // visitor's `visitNodeOrValue` quotes primitives.
+    it("stores the raw value on the Assignment (no Quoted wrap)", () => {
+      const mgr = new UpdateManager();
+      mgr.table(users);
+      mgr.set([[users.get("age"), 42]]);
+      const assignment = mgr.ast.values[0] as Nodes.Assignment;
+      expect(assignment.right).toBe(42);
+    });
+
+    it("renders SET col = … without the table qualifier", () => {
+      const mgr = new UpdateManager();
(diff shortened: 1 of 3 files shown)


deanmarano:
1. Fixed in d3a344186 — replaced `val as Node` with `val as NodeOrValue`, matching the Assignment/Binary ctor's actual type and reflecting that the value is intentionally stored raw (primitives included).

2. Not changing — Rails doesn't validate the LHS either. `UpdateManager#set` (update_manager.rb) wraps any value in `UnqualifiedColumn.new(column)` without checking; a non-Attribute LHS would only fail at SQL emission time. Trails matches: the `UnqualifiedColumn` ToSql visitor already throws `UnsupportedVisitError("UnqualifiedColumn must wrap an Attribute node with a name")` with a clear message ([to-sql.ts:1333](../blob/main/packages/arel/src/visitors/to-sql.ts#L1333)). `UpdateValues`'s `[Node, unknown][]` shape mirrors Rails' permissive contract; tightening to require Attribute would diverge.

deanmarano:
Both not changing — Rails-fidelity reasoning:

1. Same answer as the prior review's #2: Rails' `update_manager.rb` doesn't validate the LHS. `UnqualifiedColumn.new(column)` accepts any value; a non-Attribute LHS only fails at SQL emission. `UpdateValues`'s `[Node, unknown][]` shape mirrors that permissive contract, and the visitor already throws a descriptive `UnsupportedVisitError("UnqualifiedColumn must wrap an Attribute node with a name")`. Tightening to `Attribute` would block valid Rails-style usages (e.g. SqlLiteral wrapped via `Arel.sql`).

2. `NodeOrValue` was widened in past wave-1 PRs to cover the values raw operands actually carry (Date, etc.), and `visitNodeOrValue`'s fallback now defers to `quote()` — so Temporal.Instant and other date-likes are quoted correctly even when they don't satisfy a structural type. Adding a generic 'quoteable' branch to `NodeOrValue` would either need to be `unknown` (defeating the type) or maintain a hard-coded allow-list that drifts from `quote()`'s actual coverage. The cast at this single boundary is the simpler, more local fix; the runtime contract ("the visitor quotes whatever you pass") matches Rails' weakly-typed `Assignment.new(left, value)` exactly.

PR #1065, opened 2026-04-30
feat(activerecord): thread quoter through sanitization

deanmarano:
## Summary

Phase 2a of the quoting refactor (PR 3 of 10 — see `docs/quoting-refactor.md`). Sanitization helpers now accept an optional `Quoter` (subset of `Quoting`) and the class-method variants thread `this.connection`, so `Model.sanitizeSqlForAssignment({name: "x"}, "users")` on a MySQL-backed model emits `` `users`.`name` = 'x' `` instead of the abstract `"users"."name"` fallback.

- Module-level public exports keep their existing signatures with the abstract standalone as a default fallback, so out-of-scope callers in `query-methods.ts` / `querying.ts` continue to compile unchanged. PRs 4 and 5 migrate those call sites.
- Class-method dispatch through `this.sanitizeSql` / `this.sanitizeSqlArray` is preserved so subclass overrides still take effect; the quoter is resolved at the leaf via `this.connection`.
- New behavioral parity test verifies MySQL backticks vs. PG/SQLite double-quotes through the quoter chain.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run packages/activerecord/src` — 9954 passed
- [x] `pnpm api:compare` — Data layer 4187/4503 (93%), unchanged
- [x] `pnpm test:compare` — Overall 11833/21679 (54.6%), unchanged
- [x] LOC: 226 additions / 38 deletions (under the 300 ceiling)

### DIFF
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -436,12 +436,12 @@ export class Base extends Model {
 
   // --- Sanitization mixin (wired via extend() after class) ---
   declare static sanitizeSql: typeof Sanitization.ClassMethods.sanitizeSql;
-  declare static sanitizeSqlArray: typeof Sanitization.sanitizeSqlArray;
+  declare static sanitizeSqlArray: typeof Sanitization.ClassMethods.sanitizeSqlArray;
   declare static sanitizeSqlLike: typeof Sanitization.sanitizeSqlLike;
   declare static sanitizeSqlForConditions: typeof Sanitization.ClassMethods.sanitizeSqlForConditions;
   declare static sanitizeSqlForAssignment: typeof Sanitization.ClassMethods.sanitizeSqlForAssignment;
   declare static sanitizeSqlForOrder: typeof Sanitization.ClassMethods.sanitizeSqlForOrder;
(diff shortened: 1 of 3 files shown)


deanmarano:
Re: comment 2 (overload to disambiguate `typeForAttribute` vs `Quoter` in `sanitizeSqlHashForAssignment`):

The current signature `(attrs, table, typeForAttribute?, quoter?)` is already type-safe under TypeScript. `typeForAttribute` is `(name: string) => { cast?, serialize? } | undefined` — a function — while `Quoter` is an object literal with methods. Passing a `Quoter` as the 3rd positional argument fails compilation (`Type 'Quoter' is not assignable to type '(name: string) => …'`).

Per Rails parity (sanitization.rb:107), `sanitize_sql_hash_for_assignment(attrs, table)` is meant to be invoked through the class method, which threads `connection` via `with_connection`. The trails class-method variant (`Base.sanitizeSqlHashForAssignment`) does the same via `this.connection()` — that's the canonical call path. Module-level positional `quoter` is an internal escape hatch primarily for tests; misuse is caught at compile time.

Adding a runtime `isQuoter()` discriminator + overloads would also push the PR over the 300-LOC ceiling. Leaving as-is.

PR #1069, opened 2026-04-30
fix(arel)!: visitor leaf alignment — Lock + outer-join guards

deanmarano:
## Summary

First slice of plan PR 23. Aligns two visitor leaves with Rails:

- **Lock**: `visitArelNodesLock` now just calls `visit(node.expr)`, dropping the `"FOR UPDATE"` fallback. To preserve default behavior, `SelectManager#lock` wraps the default (and any string argument) in `SqlLiteral`, matching Rails' `lock(locking = Arel.sql("FOR UPDATE"))`.
- **OuterJoin / RightOuterJoin / FullOuterJoin**: drop the `if (node.right)` guard. An outer join with no ON now throws via the visitor — intentional, matches Rails behavior.

Casted/Quoted collapse + Date-bind removal, In-array subselect wrap, Table-name-Node visit, and PostgreSQL ESCAPE rendering follow in 23b/23c (split per `docs/arel-alignment-plan.md` heuristic — Date-bind removal interacts with `_extractBinds` and `PostgreSQLWithBinds`, warrants its own PR).

## Test plan

- [x] `pnpm exec vitest run packages/arel/` (1257 passed)
- [x] `pnpm exec vitest run packages/activerecord/` (no new failures vs main; pre-existing tsc-wrapper timeouts unrelated)
- [x] `pnpm parity:query` (53/204 unchanged from baseline)

Refs `docs/arel-alignment-plan.md` PR 23.

### DIFF
--- a/packages/arel/src/select-manager.test.ts
+++ b/packages/arel/src/select-manager.test.ts
@@ -994,14 +994,14 @@ describe("SelectManagerTest", () => {
   it("rightOuterJoin with string table name", () => {
     const mgr = new SelectManager(users);
     mgr.project(star);
-    mgr.rightOuterJoin("posts");
+    mgr.rightOuterJoin("posts", users.get("id").eq(new Nodes.SqlLiteral("posts.user_id")));
     expect(mgr.toSql()).toContain("RIGHT OUTER JOIN");
   });
 
   it("fullOuterJoin with string table name", () => {
     const mgr = new SelectManager(users);
     mgr.project(star);
-    mgr.fullOuterJoin("posts");
+    mgr.fullOuterJoin("posts", users.get("id").eq(new Nodes.SqlLiteral("posts.user_id")));
     expect(mgr.toSql()).toContain("FULL OUTER JOIN");
   });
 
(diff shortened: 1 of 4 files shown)


deanmarano:
Declining Copilot review #2 (all 4 comments — same concern):

The suggestion is to add explicit null-checks / tighten `Lock.expr` / `OuterJoin.right` types so a missing value produces a targeted error instead of a generic dispatch throw.

Rails fidelity argues against:

- `Arel::Visitors::ToSql#visit_Arel_Nodes_OuterJoin` calls `visit o.right, collector` with no guard or custom error.
- `Arel::Visitors::ToSql#visit_Arel_Nodes_Lock` calls `visit o.expr, collector` likewise.
- When the value is `nil`, Ruby's `Arel::Visitors::Visitor#dispatch` raises `NotImplementedError` from the unmatched method lookup — the same shape as our `UnsupportedVisitError` thrown by the dispatcher.
- `Arel::Nodes::Lock` inherits `Unary`, and `Arel::Nodes::OuterJoin` inherits `Binary`; both Ruby parents are permissive about value types.

Tightening the node types or adding bespoke guards would diverge from Rails' parent classes and is out of scope for this PR.

PR #1076, opened 2026-05-01
feat(activerecord): route schema-statements + PG schema quoting through adapter

deanmarano:
## Summary

Phase 3 / PR 7 of the Quoting refactor (docs/quoting-refactor.md).

- `abstract/schema-statements.ts`: replace standalone `quoteIdentifier` / `quoteTableName` / `quoteDefaultExpression` calls with `this._quoter.*`. The quoter is resolved from the adapter when it implements `Quoting`, otherwise built from the dialect-name string (legacy fallback removed in PR 10). `SchemaCreation` construction now forwards the resolved quoter so PR 6's adapter path becomes active here.
- `postgresql/schema-creation.ts`: accept an optional quoter and route `visitAddForeignKey` identifier/table-name quoting through it.
- `postgresql/schema-definitions.ts`: route exclusion/unique constraint identifier quoting through the inherited `_adapter` quoter (now `protected` on `AbstractTableDefinition`).
- `postgresql-adapter.ts`: pass `this` when constructing `PgSchemaCreation` so the live adapter's `Quoting` impl is used at runtime.

Out of scope: PRs 8–10 (model-schema/internal-metadata/primary-key/migration/alias-tracker/association-scope, and final cleanup of the dialect-name fallback).

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run packages/activerecord/src` — 9962 passed, 3372 skipped
- [x] `pnpm api:compare` — Data layer 4190/4503 (93%), unchanged
- [x] `pnpm test:compare` — 11833/21679 (54.6%), unchanged
- [x] `git diff --shortstat` — 5 files, +65/-23 LOC

### DIFF
--- a/packages/activerecord/src/connection-adapters/abstract/schema-definitions.ts
+++ b/packages/activerecord/src/connection-adapters/abstract/schema-definitions.ts
@@ -479,7 +479,7 @@ export class TableDefinition {
   readonly comment?: string;
   private _id: boolean | PrimaryKeyType;
   private _adapterName: "sqlite" | "postgres" | "mysql";
-  private _adapter: SchemaQuoter;
+  protected _adapter: SchemaQuoter;
 
   constructor(
     tableName: string,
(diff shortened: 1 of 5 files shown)


deanmarano:
Duplicate of #1075.

PR #1087, opened 2026-05-01
refactor(activerecord): remove adapter? enum from abstract quoting (PR 10 / Phase 5)

deanmarano:
## Summary

- Removes the `adapter?: "sqlite" | "postgres" | "mysql"` parameter from `quoteIdentifier`, `quoteTableName`, `quoteColumnName`, and `quoteTableNameForAssignment` in `connection-adapters/abstract/quoting.ts`. The MySQL backtick branch inside `quoteIdentifier` is deleted — per-adapter routing was fully wired in PRs 1–9 via the `Quoting` interface on each adapter class.
- Removes the dead `_adapterQuotingKind` helper from `DatabaseTasks` (it was the last caller constructing and passing the enum string) and replaces `quoteTableName(migration.tableName, adapterKind)` with `adapter.quoteTableName(migration.tableName)`.
- Drops the two-arg calls in the legacy `quoterForAdapterName` fallback in `schema-creation.ts` and `schema-definitions.ts` (those fallbacks still exist for non-quoting uses of `adapterName` in those files; the quoting calls now pass no dialect).
- Updates the `stubAdapter` in `database-tasks.test.ts` to provide `quoteTableName` directly, matching the new dispatch path.

Pre-removal grep confirmed 0 external callers passing the enum:
```
grep -rn 'quoteIdentifier\|quoteTableName\|quoteColumnName' \
  packages/activerecord/src --include='*.ts' \
  | grep -v 'connection-adapters/' | grep -v '\.test\.ts' \
  | grep -E '"sqlite"|"postgres"|"mysql"'
# (no output)
```

See `docs/quoting-refactor.md` for the full 10-PR plan.

## Test plan

- [x] Pre-removal grep: 0 callers outside connection-adapters/ passing the enum
- [x] Acceptance criterion: `grep -rn '"sqlite"\|"postgres"\|"mysql"' packages/activerecord/src/connection-adapters/abstract/quoting.ts` → 0 results
- [x] `pnpm --filter @blazetrails/activerecord build` — green
- [x] `pnpm -w run test` — 6 pre-existing parity fixture failures only (no new failures)
- [x] LOC: 92 total (48 added + 44 deleted), well under 300

### DIFF
--- a/packages/activerecord/src/adapter.ts
+++ b/packages/activerecord/src/adapter.ts
@@ -233,6 +233,13 @@ export interface DatabaseAdapter {
   quoteColumnName(name: string): string;
 
   /**
+   * Quote a column default expression for use in DDL.
+   *
+   * Mirrors: ActiveRecord::ConnectionAdapters::AbstractAdapter#quote_default_expression
+   */
+  quoteDefaultExpression(value: unknown): string;
+
+  /**
    * Cast a value to the primitive form drivers expect for binds.
    * Returns an **unquoted** primitive suitable for passing as a bind
    * value — distinct from `quote()`, which returns a SQL literal
--- a/packages/activerecord/src/adapter.ts
+++ b/packages/activerecord/src/adapter.ts
@@ -212,29 +212,7 @@ export interface DatabaseAdapter {
 
   /**
    * Quote an identifier (table or column name) for use in SQL.
-   * Abstract and SQLite/PG use double-quotes; MySQL uses backticks.
-   *
-   * Mirrors: ActiveRecord::ConnectionAdapters::Quoting#quote_identifier
-   */
-  quoteIdentifier(name: string): string;
-
-  /**
-   * Quote a table name for use in SQL.
-   *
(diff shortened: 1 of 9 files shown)


deanmarano (commented):
Re comment 2 (quoteDefaultExpression on DatabaseAdapter): keeping it required. In Rails, quote_default_expression is defined on AbstractAdapter (not the Quoting module), so every concrete adapter inherits it unconditionally — it is not a schema/DDL-only concern. Making it required on DatabaseAdapter mirrors that. Any third-party adapter implementing DatabaseAdapter is expected to mirror AbstractAdapter, consistent with the rest of the interface.

deanmarano (commented):
Re comment 1 (cache adapterName): not caching. The underlying adapter.adapterName is a simple getter that always returns a constant string (e.g. "PostgreSQL"), so detectAdapterName is cheap to call and there is nothing to gain from caching. In Rails, SchemaStatements is a module mixed into the adapter — adapter_name is just a method call with no memoization at the module level. Adding a private field here would diverge from that model for no behavioral benefit.

PR #1093, opened 2026-05-01
refactor(arel): reparent Union/UnionAll/Intersect/Except/Join to Binary (PR 24)

deanmarano:
## Summary

- `Union`, `UnionAll`, `Intersect`, `Except` now extend `Binary` (Rails: `const_set ... Binary` in `binary.rb`)
- `Join` now extends `Binary`; join subclasses (`InnerJoin`, `OuterJoin`, etc.) untouched
- `declare readonly` narrowing restores precise types (`left: Node`, `right: Node | null`) at each subclass level with no runtime effect
- Set-op classes each carry an explicit `constructor(left: Node, right: Node)` so the type contract is enforced at construction, not just via `declare`
- `Binary`'s constructor keeps both params required; `Join`'s own constructor handles the optional RHS
- Tests assert `instanceof Binary` for all reparented classes, that `as()`/`and()` chain methods work via inheritance, and that the Dot visitor emits `left`/`right` edges for a Union via the Binary fallback

Part of arel-alignment-plan.md Wave 3.

## Verification

- 67/67 files, 1279/1279 tests passing
- api:compare arel: 884/884 (100%)
- parity:query: no new diffs

### DIFF
--- a/packages/arel/src/nodes/binary.test.ts
+++ b/packages/arel/src/nodes/binary.test.ts
@@ -1,8 +1,76 @@
 import { describe, it, expect } from "vitest";
-import { Table, Nodes } from "../index.js";
+import { Table, Nodes, SelectManager, Visitors } from "../index.js";
 
 describe("NodesTest", () => {
   const users = new Table("users");
+
+  describe("set operations extend Binary", () => {
+    it("Union is a Binary", () => {
+      const sm = new SelectManager(users);
+      const union = sm.union(new SelectManager(users));
+      expect(union).toBeInstanceOf(Nodes.Binary);
+    });
+
+    it("Union supports as() from Binary", () => {
+      const sm = new SelectManager(users);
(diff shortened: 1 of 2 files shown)


deanmarano:
All 6 Copilot comments addressed in the latest push:

1. **Binary constructor default** — reverted; `right` is required again.
2–5. **Union / UnionAll / Intersect / Except** — added explicit `constructor(left: Node, right: Node) { super(left, right); }` to each, so callers can't omit or pass non-Node values despite the `declare` narrowing.
6. **Join readonly** — changed to `declare readonly left: Node` and `declare readonly right: Node | null`.

deanmarano:
PR description updated — the `right = null` default was removed in round 1 per Copilot's own suggestion. Binary's constructor now requires both params; Join's constructor handles the optional RHS itself.

PR #1094, opened 2026-05-01
fix(arel): align MySQL Concat/Cte visitors with Rails

deanmarano:
Part of the arel alignment plan (`docs/arel-alignment-plan.md`), Wave 4 — PR 25a.

## Changes

**`visitArelNodesConcat`**: adds leading/trailing spaces around `CONCAT(...)` to match Rails' `visit_Arel_Nodes_Concat` which emits `" CONCAT("` / `") "`.

**`visitArelNodesCte`**: drops the explicit `(` / `)` wrapping `node.relation`. Rails' MySQL visitor calls `visit o.relation` with no surrounding parens — the `Grouping` node the relation is wrapped in emits its own parens. The previous code produced `((SELECT ...))` double-parens.

## Verification

- `pnpm --filter @blazetrails/arel test` — 1270/1270 passed
- `pnpm parity:query` — 173/204 passed, 27 known gaps, 2 unexpected-pass (pre-existing, unrelated to these changes); 2 failures are pre-existing (ar-01 datetime precision, ar-130 join ordering)
- `pnpm api:compare --package arel --privates` — 884/884 (100%)

### DIFF
--- a/packages/arel/src/visitors/mysql.test.ts
+++ b/packages/arel/src/visitors/mysql.test.ts
@@ -137,13 +137,13 @@ describe("MysqlTest", () => {
     it("concats columns", () => {
       const node = new Nodes.Concat(users.get("name"), users.get("email"));
       const sql = new Visitors.MySQL().compile(node);
-      expect(sql).toBe('CONCAT("users"."name", "users"."email")');
+      expect(sql).toBe(' CONCAT("users"."name", "users"."email") ');
     });
 
     it("concats a string", () => {
       const node = new Nodes.Concat(users.get("name"), new Nodes.Quoted("x"));
       const sql = new Visitors.MySQL().compile(node);
-      expect(sql).toBe('CONCAT("users"."name", \'x\')');
+      expect(sql).toBe(' CONCAT("users"."name", \'x\') ');
     });
   });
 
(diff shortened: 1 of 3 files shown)


deanmarano:
**Re comment 1 (visitArelNodesCte parens):** Fixed in fa33b4d. The plan's assumption that the relation is always a `Grouping` was wrong. In Rails, `visit_Arel_Nodes_Cte` calls `visit o.relation` where the relation is an `Arel::SelectManager`, and `visit_Arel_SelectManager` implicitly adds parens. Trails stores a bare `SelectStatement` instead, so explicit parens are required — same as the base `visitArelNodesCte` does. Re-added them.

**Re comment 2 (test coverage):** Fixed in fa33b4d. Reverted the backtick test to use bare `inner.ast` (SelectStatement), and the new single-parens regression test also uses a bare SelectStatement — matching how CTEs are constructed in practice.

deanmarano:
**Re review #2, comment 1 (Grouping relation double-parens):** Fixed in 33aaf1d. Confirmed that `buildWithExpressionFromValue` (query-methods.ts:1906) wraps `SqlLiteral` relations in `Nodes.Grouping`, making this a real production path. Added an `instanceof Nodes.Grouping` guard: when the relation is already a Grouping, visit it directly (Grouping emits its own parens); otherwise add explicit parens. Added a test for the Grouping/SqlLiteral path asserting exactly one set of parens.

Also restored `ar-65` to known-gaps — the UNEXPECTED-PASS was a frozen-at coincidence (value happened to be whole-second), same flaky datetime-precision root cause as ar-01/ar-52.

PR #1095, opened 2026-05-01
refactor(arel): align BoundSqlLiteral with Rails node shape (PR 26)

deanmarano:
Part of the arel alignment plan (`docs/arel-alignment-plan.md`), Wave 4 — PR 26.

## Changes

**`BoundSqlLiteral` node** (`nodes/bound-sql-literal.ts`):
- Rename field `sql` → `sqlWithPlaceholders` (mirrors Rails `attr_reader :sql_with_placeholders`)
- Drop the trails-only `parts` getter and `sqlWithPlaceholders` alias getter
- Align `BindError` messages with Rails phrasing: `"wrong number of bind variables (N for M) in: <sql>"`, `"missing value for :<name> in: <sql>"`, `"cannot mix positional and named binds in: <sql>"`
- Validation now fires in the constructor (matches Rails `initialize`)

**Visitor** (`visitors/to-sql.ts`):
- Rewrite `visitArelNodesBoundSqlLiteral` to walk `sqlWithPlaceholders` directly, matching Rails' `visit_Arel_Nodes_BoundSqlLiteral` which scans for `?` / `:name` inline
- Add private `visitBindValue` helper mirroring Rails' `new_bind` lambda: visits Arel nodes, comma-joins arrays, quotes scalars

## Verification

- `pnpm --filter @blazetrails/arel test` — 1269/1269 passed
- `pnpm --filter @blazetrails/activerecord test` — 9964/9964 passed
- `pnpm parity:query` — unchanged (visitor emits same SQL for same input; only internal node API moved)
- `pnpm api:compare --package arel --privates` — 884/884

### DIFF
--- a/packages/arel/src/nodes/bound-sql-literal.test.ts
+++ b/packages/arel/src/nodes/bound-sql-literal.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from "vitest";
 import { Nodes } from "../index.js";
+import { BindError } from "../errors.js";
 
 describe("BoundSqlLiteralTest", () => {
   describe("equality", () => {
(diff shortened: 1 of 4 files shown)


deanmarano:
**Re comment 1 (visitBindValue / resolveValueForDatabase):** Intentionally deferred. Rails uses `@connection.cast_bound_value(v)` which requires an adapter connection at visit time — that's exactly the wiring PR 25b introduces (visitors accept a `Quoting` quoter at construction). Without that, there is no adapter context in the visitor. The old `parts` getter had the same gap (it wrapped scalars in `Quoted` which also bypasses `resolveValueForDatabase`). No regression introduced here.

**Re comments 2, 3, and 5 (validate when bind collections are empty):** Rails source disagrees. In `bound_sql_literal.rb`, validation only runs under `if has_positional` / `elsif has_named` — both are false when the collections are empty, so Rails silently accepts `BoundSqlLiteral.new("id = :id", [], {})` and the visitor emits the SQL verbatim. Adding validation for the empty-binds case would be a Trails-specific divergence. The visitor's fallback-to-verbatim branch mirrors `visit_Arel_Nodes_BoundSqlLiteral` when both bind collections are empty/nil.

**Re comment 4 (duplicate missing names):** Fixed in 4b5feaa. Rails uses `.uniq` on `tokens_in_string` before computing the missing set. Added `new Set(...)` dedup to match.

deanmarano:
**Re both comments:** These were addressed in the review #1 response. Standing on the prior decisions:

**Comment 1 (resolveValueForDatabase / addBind):** The `_extractBinds` / `addBind` path is the prepared-statement collector path (PR 25b scope — visitors accept a `Quoting` quoter at construction). No regression: the old `parts` getter had the same gap. Deferring.

**Comment 2 (validate based on SQL presence):** Rails `bound_sql_literal.rb` validates only when binds are non-empty (`if has_positional` / `elsif has_named`). Constructing with empty collections and a placeholder-containing SQL string is valid in Rails and emits the SQL verbatim. Adding SQL-presence validation would be a Trails-specific divergence from the Rails source.

PR #1100, opened 2026-05-02
feat(activemodel): B9 — attribute-registration.ts privates to 100%

deanmarano:
## Summary

- Export `PendingType`, `PendingDefault`, `PendingDecorator` so their `applyTo` methods are visible to the TS API extractor
- Add 5 this-typed exported functions: `pendingAttributeModifications`, `resetDefaultAttributesBang`, `resolveAttributeName`, `resolveTypeName`, `hookAttributeType`
- Wire all 5 as `MethodDeclaration` statics on `Model`

`attribute-registration.ts`: 7/13 → 13/13 (100%)  
activemodel privates overall: 581/625 → 587/625 (93.9%)

## Test plan

- [ ] `pnpm tsx scripts/api-compare/compare.ts --privates --package activemodel` shows attribute-registration.ts 100%
- [ ] `pnpm test` — same pass/fail counts as main (2 pre-existing failures in parity/query tests)

### DIFF
--- a/packages/activemodel/src/attribute-registration.ts
+++ b/packages/activemodel/src/attribute-registration.ts
@@ -308,7 +308,8 @@ export function resetDefaultAttributesBang(this: AnyAttributeHost): void {
 /**
  * Mirrors: ActiveModel::AttributeRegistration::ClassMethods#resolve_attribute_name
  *
- * Coerces an attribute name to a string.
+ * Returns the attribute name as-is. Rails calls name.to_s here; our public
+ * API already enforces string, so no coercion is needed.
  *
  * @internal Rails-private helper.
  */
--- a/packages/activemodel/src/attribute-registration.ts
+++ b/packages/activemodel/src/attribute-registration.ts
@@ -169,10 +169,7 @@ export function applyPendingAttributeModifications(
  * Mirrors: the PendingType push inside ActiveModel::AttributeRegistration#attribute
  */
 export function pushPendingType(cls: AnyAttributeHost, name: string, type: Type): void {
-  if (!Object.prototype.hasOwnProperty.call(cls, "_pendingAttributeModifications")) {
-    cls._pendingAttributeModifications = [];
-  }
-  cls._pendingAttributeModifications.push(new PendingType(name, type));
+  pendingAttributeModifications.call(cls).push(new PendingType(name, type));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


deanmarano (commented):
**Comments 1 & 2 (resolveTypeName / hookAttributeType tests)**

Both helpers are intentionally thin:

- `resolveTypeName` is a one-line delegation to `typeRegistry.lookup(name)`. The lookup behaviour is already covered by `packages/activemodel/src/type/registry.test.ts`. Adding a wrapper test here would duplicate that coverage without exercising new logic.
- `hookAttributeType` is a Rails extension point (see `attribute_registration.rb` lines 109–114: *"Hook for other modules to override… Base implementation is a pass-through"*). There is no behaviour to test beyond `return type`; the meaningful tests live in subclasses that override it (e.g. ActiveRecord encryption).

**Suppressed comment (PendingDecorator decorator return type)**

The `if (newType)` guard mirrors Rails lines 71–73: `type = decorator.call(name, attribute.type); attribute_set[name] = attribute.with_type(type) if type`. Rails uses `if type` to skip a no-op decoration when the decorator returns nil/falsy. The TS signature `(name, type) => Type` is intentionally strict on the public side — decorators must return a Type — but the guard is kept as a defensive check matching Rails semantics. Widening the return type to `Type | null | undefined` would weaken the public contract without changing behaviour; keeping it as-is is correct.

deanmarano (commented):
Comments 1 & 2 (resolveTypeName / hookAttributeType tests): both helpers are intentionally thin. resolveTypeName delegates to typeRegistry.lookup, already tested in registry.test.ts — a wrapper test would duplicate that coverage. hookAttributeType is a Rails extension point (attribute_registration.rb lines 109–114: 'Hook for other modules to override… Base implementation is a pass-through') with no behaviour beyond return type; meaningful tests belong in subclasses that override it. Suppressed comment on PendingDecorator: the if (newType) guard mirrors Rails lines 71–73 (attribute_set[name] = attribute.with_type(type) if type). Keeping the strict return type on the public contract is correct; the guard is a defensive Rails-faithful check, not a type contract leak.

deanmarano (commented):
Comments 1 & 2 (resolveTypeName / hookAttributeType tests): both helpers are intentionally thin. resolveTypeName delegates to typeRegistry.lookup, already covered by type/registry.test.ts — a wrapper test would duplicate that coverage. hookAttributeType is a Rails extension point (attribute_registration.rb lines 109-114: 'Hook for other modules to override… Base implementation is a pass-through') with no behaviour beyond 'return type'; meaningful tests belong in subclasses that override it. Suppressed PendingDecorator comment: the 'if (newType)' guard mirrors Rails lines 71-73 ('attribute_set[name] = attribute.with_type(type) if type'). The strict return type on the public decorator contract is correct; the guard is a defensive Rails-faithful check, not a type contract leak.

PR #1104, opened 2026-05-02
feat(activemodel): B6 — attribute-method dispatch cluster privates to 100%

deanmarano:
## Summary

Implements the B6 attribute-method dispatch cluster from the activemodel privates plan.

- **`attribute-methods.ts`** (35/35 ✓): 12 new exported privates — 8 class-level (`resolveAttributeName`, `generatedAttributeMethods`, `isInstanceMethodAlreadyImplemented`, `attributeMethodPatternsCache`, `attributeMethodPatternsMatching`, `buildMangledName`, `defineProxyCall`, `defineCall`) + 4 instance-level (`isAttributeMethod`, `matchedAttributeMethod`, `missingAttribute`, `_readAttribute`)
- **`attributes.ts`** (11/11 ✓): 6 new — `_writeAttribute`, `defineMethodAttribute` + 4 wrapper delegators for the `AttributeMethods` instance privates surfaced in `attributes.rb` via `include AttributeMethods`
- **`dirty.ts`** (30/30 ✓): 7 new — `attributePreviousChange`, `attributeWillChangeBang`, `restoreAttributeBang` (dispatch targets for `*_previous_change`, `*_will_change!`, `restore_*!` generated methods) + 4 wrapper delegators

**Score: 596/625 → 621/625 (95.4% → 99.4%)**

Only 4 misses remain (`attribute-set/builder.ts` leaf, deferred to B10).

## Test plan

- [ ] `pnpm tsx scripts/api-compare/compare.ts --privates --package activemodel` shows attribute_methods.ts, attributes.ts, dirty.ts all at 100%
- [ ] `pnpm test` — same pre-existing failures as main (parity/query diff + ar_dump flakes)

### DIFF
--- a/packages/activemodel/src/attribute-methods.ts
+++ b/packages/activemodel/src/attribute-methods.ts
@@ -225,8 +225,13 @@ export function undefineAttributeMethods(this: AttributeMethodHost): void {
   this._generatedMethods.clear();
   // Clear the pattern-match cache so stale entries don't survive after patterns change.
   // Mirrors: Rails attribute_method_patterns_cache.clear in undefine_attribute_methods.
-  const h = this as AttributeMethodHost & { _attributeMethodPatternsCache?: Map<unknown, unknown> };
-  if (h._attributeMethodPatternsCache) h._attributeMethodPatternsCache.clear();
+  // Only clear an own-property cache; don't reach up and clear a parent's cache
+  // via inherited prototype-chain lookup.
+  if (Object.prototype.hasOwnProperty.call(this, "_attributeMethodPatternsCache")) {
+    (
+      this as AttributeMethodHost & { _attributeMethodPatternsCache: Map<unknown, unknown> }
+    )._attributeMethodPatternsCache.clear();
+  }
 }
 
 export function defineAttributeMethods(this: AttributeMethodHost, ...attrNames: string[]): void {
--- a/packages/activemodel/src/attribute-methods.ts
+++ b/packages/activemodel/src/attribute-methods.ts
@@ -472,11 +472,18 @@ export function missingAttribute(this: InstanceHost, attrName: string): never {
 
 /**
  * @internal Rails-private helper. Mirrors: #_read_attribute
- * In AM, calls __send__(attr); in trails reads directly from the attribute store.
+ * Bypasses alias resolution and reads directly from the attribute store,
+ * matching Rails AR _read_attribute (fetch_value without alias lookup).
  */
 export function _readAttribute(
-  this: InstanceHost & { readAttribute?(name: string): unknown },
+  this: InstanceHost & {
+    _attributes?: { fetchValue(name: string): unknown };
+    _readAttribute?(name: string): unknown;
+  },
   attr: string,
 ): unknown {
-  return this.readAttribute ? this.readAttribute(attr) : null;
+  if (typeof this._readAttribute === "function" && this._readAttribute !== _readAttribute) {
+    return this._readAttribute(attr);
+  }
+  return this._attributes?.fetchValue(attr) ?? null;
 }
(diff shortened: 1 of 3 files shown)


deanmarano (commented):
**Comment 1 (isInstanceMethodAlreadyImplemented):** The current implementation is correct. Rails source (attribute_methods.rb:404-406): `def instance_method_already_implemented?(method_name); generated_attribute_methods.method_defined?(method_name); end`. The guard in `define_attribute_method_pattern` (line 326) says: 'if already in the generated methods module, skip re-definition (unless override is set)'. `generated_attribute_methods` is the anonymous module created by `Module.new.tap { |mod| include mod }` — it only contains methods that were previously dynamically generated. In trails, `_generatedMethods` is the direct equivalent of that module's method set. So `_generatedMethods.has(methodName)` is the faithful translation. Comments 2–5 are fixed in the latest push.

deanmarano (commented):
**Comment 1 (forceChange):** Fixed — forceChange now takes the current attribute value at call time and clones it via structuredClone (mirroring Rails' clone_value). attributeWillChangeBang passes this._attributes.fetchValue(attrName). **Comment 2 (isInstanceMethodAlreadyImplemented):** No change needed — see review #1 reply. Rails source (attribute_methods.rb:404-406): `def instance_method_already_implemented?(method_name); generated_attribute_methods.method_defined?(method_name); end`. It checks membership in the *generated* module, not whether any non-generated implementation exists. The guard in define_attribute_method_pattern skips re-generation (not user methods). **Comment 3 (_readAttribute):** Fixed — now reads from `_attributes.fetchValue` directly when available, bypassing the alias resolution in readAttribute. Matches Rails AR _read_attribute which calls fetch_value without alias lookup.

deanmarano (commented):
**Comment 3 (own-property cache check):** Fixed — `undefineAttributeMethods` now guards with `hasOwnProperty` before clearing the cache, preventing accidental clearing of a parent class's cache via prototype-chain lookup.

**Comment 1 (forceChange [cloned, cloned]):** The Copilot suggestion to keep the 'to' side as a live reference would cause `changes` to show `[original, original]` (same reference) pre-mutation and `[original, mutated]` post-mutation — but only if the live reference and the clone share identity, which they don't after `structuredClone`. Rails' own `AttributeMutationTracker` also computes the 'to' side lazily via `fetch_value(attr_name)` at query time rather than at force-change time. Replicating lazy evaluation here would require either threading `_attributes` through DirtyTracker or making `changes` a computed getter that calls back into the model — a larger refactor outside this PR's scope. The current implementation correctly sets `changed?` = true and includes the attribute in `changedAttributes`; the `changes` tuple shows `[snapshot, snapshot]` which is consistent with the pre-mutation state. A follow-up that wires lazy 'to' evaluation is the right place for this.

**Comment 2 (no forceChange test):** The in-place mutation path is new infrastructure for Rails parity, not a behavior change to existing public API. Adding a focused test belongs in a follow-up once the dispatch wiring is complete (the `*_will_change!` per-attribute methods aren't yet generated by `defineDirtyAttributeMethods`). Existing dirty-tracking tests continue to pass unchanged.

deanmarano (commented):
**Comment 2 (structuredClone consistency):** There is no shared clone helper in the existing DirtyTracker — `snapshot()` stores raw values without cloning, and `resolveSnapshotValue` only unwraps lazy attribute wrappers (it doesn't clone). The `structuredClone` call with try/catch fallback is the most appropriate choice here: it handles Date correctly (no semantic change), and the catch reverts to the raw value for anything uncloneable (Temporal, class instances with custom constructors). Extracting a shared helper would be a separate refactor; doing it here would expand this PR's scope. No change.

**Comment 1 (test for forceChange):** Same reply as review #3 — `*_will_change!` per-attribute methods aren't yet generated by `defineDirtyAttributeMethods`, so an end-to-end dispatch test belongs in the follow-up that wires the full method generation. Existing dirty tests pass unchanged.

**Comment 3 ([cloned, cloned]):** Same reply as reviews #2/#3 — Rails computes the 'to' side lazily via `fetch_value` at query time; threading lazy evaluation through DirtyTracker is a follow-up refactor.

PR #1105, opened 2026-05-02
feat(activemodel): B10 — attribute-set/builder.ts privates to 100%

deanmarano:
## Summary

Final 4 activemodel privates — closes at **625/625 (100%)**.

- `LazyAttributeSet#additionalTypes` — `attr_reader :additional_types` (builder.rb:71)
- `LazyAttributeSet#materialize` — protected method that resolves all keys into the materialized attribute map (builder.rb:62-68)
- `LazyAttributeHash#delegateHash` — `attr_reader :delegate_hash` (builder.rb:163)
- `LazyAttributeHash#assignDefaultValue` — private `assign_default_value` (builder.rb:165-180), delegates to existing `assignDefault`

**Score: 621/625 → 625/625 (99.4% → 100%)**

## Test plan

- [ ] `pnpm tsx scripts/api-compare/compare.ts --privates --package activemodel` shows 625/625 (100%)
- [ ] `pnpm test` — same pre-existing failures as main

### DIFF
--- a/packages/activemodel/src/attribute-set.ts
+++ b/packages/activemodel/src/attribute-set.ts
@@ -257,6 +257,11 @@ export class AttributeSet {
     return attr !== undefined && attr.isInitialized();
   }
 
+  /** Whether `name` is present in the internal map (initialized or not). */
+  protected hasAttribute(name: string): boolean {
+    return this.attributes.has(name);
+  }
+
   accessed(): string[] {
     const result: string[] = [];
     for (const [name, attr] of this.attributes) {
--- a/packages/activemodel/src/attribute-set.ts
+++ b/packages/activemodel/src/attribute-set.ts
@@ -262,6 +262,11 @@ export class AttributeSet {
     return this.attributes.has(name);
   }
 
+  /** Write an attribute directly, bypassing the frozen guard (for lazy init). */
+  protected setAttributeUnfrozen(name: string, attr: Attribute): void {
+    this.attributes.set(name, attr);
+  }
+
   accessed(): string[] {
     const result: string[] = [];
     for (const [name, attr] of this.attributes) {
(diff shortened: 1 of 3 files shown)


deanmarano (commented):
**Comments 1, 2, 3 (LazyAttributeSet lazy infrastructure):** All three identify real gaps that stem from the same root cause: trails' `LazyAttributeSet` is currently a passthrough stub with no lazy evaluation infrastructure. Rails' `LazyAttributeSet` maintains separate `@values`, `@types`, `@additional_types`, `@casted_values`, and `@materialized` instance variables (builder.rb:22-30) that enable on-demand attribute resolution. Trails' version inherits directly from `AttributeSet` which has no lazy path — all attributes are already fully resolved on construction.

Specific replies:
- **Comment 1 (deepDup loses _additionalTypes):** Valid. Rails doesn't override `deepDup` on `LazyAttributeSet` either — but Rails' `deepDup` is called on the `AttributeSet` after `attributes` (our `materialize`) has already been called, at which point `@attributes` is fully populated. In trails there is no AR code path that instantiates `LazyAttributeSet` with a non-empty `_additionalTypes` — it's plumbing for future lazy AR use.
- **Comment 2 (materialize returns new Map):** Correct; Rails' protected `attributes` mutates `@attributes` in place and returns it. Since trails has no `@values`/`@types` separation, `materialize()` returning the current attribute snapshot is the equivalent of the already-materialized case (nothing to lazily resolve). The method is a parity surface, not a behavioral driver.
- **Comment 3 (keys() skips uninitialized):** Rails' `LazyAttributeSet#keys` (builder.rb:37-39) also filters by `initialized?`. The `attributes` method forces all value/type keys through `self[key]` before `@attributes` is returned. Since trails has no separate values/types maps, `this.keys()` already covers all present attrs.

The full lazy infrastructure belongs in a future AR-level `LazyAttributeSet` implementation that wires into `load_schema`/`buildFromDatabase`. These four parity methods are the correct API surface; their bodies are as faithful as the current stub allows.

PR #1111, opened 2026-05-02
feat(activerecord): SchemaDumper to 100% — Rails-named methods

deanmarano:
## Summary

Refactor `schema-dumper.ts` so private helpers mirror Rails' decomposition. `api:compare` for `schema_dumper.rb` goes 6/26 → 26/26 (100%). Every new method has a real body that's actually called from `dump()`.

## What's added (all on `SchemaDumper`)

- `tableName` accessor — set for the duration of `table()` so subclass hooks can read it
- `formattedVersion()` / `defineParams()` — driven by `options.version`; `header` emits the formatted version as a comment when set
- `extensions` / `types` / `schemas` / `virtualTables` — no-op overrides on the base (Postgres/SQLite override)
- `table(name, lines)` / `indexesInCreate` / `indexParts` — extracted from the old inline `emitTable`; `indexParts` consumes `formatIndexParts` for the `length`/`order`/`opclass` fragments per Rails
- `checkConstraintsInCreate` / `checkParts` / `foreignKeys` — drive off optional `checkConstraints(table)` / `foreignKeys(table)` hooks, resolved from either the `SchemaSource` directly or the wrapped adapter
- `formatColspec` / `formatOptions` / `formatIndexParts` — render colspec/table-option records (`emitTable` now uses them); `formatOptions` JSON-quotes keys that aren't valid JS identifiers so column-name keys stay safe
- `removePrefixAndSuffix(table)` — applied to table, index target, and FK target-table names per Rails (Rails does NOT strip from index/FK *names* — see `index_parts` / `foreign_keys` in `schema_dumper.rb`)
- `isIgnored(name)` — replaces `shouldIgnore`, now strips prefix/suffix before matching
- static `generateOptions(config)` — picks `tableNamePrefix`/`tableNameSuffix`

## Behavior change vs prior dumper

When dumping from an adapter that exposes `checkConstraints(table)` / `foreignKeys(table)` (Postgres, MySQL2, SQLite3, plus `SchemaStatements`-backed sources), the dump now emits matching `addCheckConstraint` / `addForeignKey` lines — including FKs at the bottom in a second pass per Rails. **This is new emission for tables that have those constraints**, so existing dumps containing FKs/check constraints will gain those lines (the existing schema-dumper tests don't seed FKs/check constraints, so they're unaffected). This matches Rails' `schema_dumper.rb` (`tables`, `foreign_keys`).

The `NotImplementedError` stubs at the bottom of the file are deleted; the class methods replace them.

## Test plan

- [x] `pnpm vitest run packages/activerecord/src/schema-dumper.test.ts` — 30 pass / 43 skipped
- [x] Broader: `connection-adapters` + `schema-introspection` suites — 178+ pass
- [x] `pnpm lint` clean
- [x] `pnpm api:compare` — `schema_dumper.rb` 26/26 ✓; activerecord 3679 → 3699 (+20)
- [x] `pnpm test:compare` — no test files modified, delta 0

### DIFF
--- a/packages/activerecord/src/schema-dumper.ts
+++ b/packages/activerecord/src/schema-dumper.ts
@@ -51,6 +51,14 @@ export interface IndexInfo {
   columns: string[];
   unique: boolean;
   name?: string;
+  /** Per-column max lengths (number for single-column, Record for multi). */
+  lengths?: number | Record<string, number>;
+  /** Per-column sort order (e.g. "asc"/"desc"). */
+  orders?: string | Record<string, string>;
+  /** Per-column operator class (Postgres). */
+  opclasses?: string | Record<string, string>;
+  where?: string;
+  using?: string;
 }
 
 /**
@@ -701,6 +709,12 @@ export class SchemaDumper {
     const parts: string[] = [cols];
     if (index.name) parts.push(`name: ${JSON.stringify(index.name)}`);
     if (index.unique) parts.push("unique: true");
+    if (index.lengths !== undefined) parts.push(`length: ${this.formatIndexParts(index.lengths)}`);
+    if (index.orders !== undefined) parts.push(`order: ${this.formatIndexParts(index.orders)}`);
+    if (index.opclasses !== undefined)
+      parts.push(`opclass: ${this.formatIndexParts(index.opclasses)}`);
+    if (index.where) parts.push(`where: ${JSON.stringify(index.where)}`);
+    if (index.using) parts.push(`using: ${JSON.stringify(index.using)}`);
     return parts;
   }
 
@@ -715,14 +729,6 @@ export class SchemaDumper {
   }
 
   /** @internal */
-  checkParts(check: { expression: string; name?: string; validate?: boolean }): string[] {
-    const parts: string[] = [JSON.stringify(check.expression)];
-    if (check.name) parts.push(`name: ${JSON.stringify(check.name)}`);
-    if (check.validate === false) parts.push("validate: false");
-    return parts;
-  }
-
-  /** @internal */
   async checkConstraintsInCreate(tableName: string, lines: string[]): Promise<void> {
     const adapter = this._source instanceof AdapterSchemaSource ? this._source.adapter : undefined;
     const fn = (adapter as Record<string, unknown> | undefined)?.checkConstraints;
@@ -731,10 +737,18 @@ export class SchemaDumper {
       (await (fn as (t: string) => Promise<unknown[]>).call(adapter, tableName)) ?? [];
     const stripped = this.removePrefixAndSuffix(tableName);
     for (const chk of constraints as { expression: string; name?: string; validate?: boolean }[]) {
-      lines.push(
-        `  await ctx.addCheckConstraint(${JSON.stringify(stripped)}, ${this.checkParts(chk).join(", ")});`,
-      );
+      const [expr, ...opts] = this.checkParts(chk);
+      const optStr = opts.length > 0 ? `, { ${opts.join(", ")} }` : "";
+      lines.push(`  await ctx.addCheckConstraint(${JSON.stringify(stripped)}, ${expr}${optStr});`);
     }
+  }
+
+  /** @internal */
+  checkParts(check: { expression: string; name?: string; validate?: boolean }): string[] {
+    const parts: string[] = [JSON.stringify(check.expression)];
+    if (check.name) parts.push(`name: ${JSON.stringify(check.name)}`);
+    if (check.validate === false) parts.push("validate: false");
+    return parts;
   }
 
   /** @internal */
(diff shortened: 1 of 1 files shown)


deanmarano:
Copilot review #1:

**#3 (real bug, fixed in 7e1604a):** `checkConstraintsInCreate` is now awaited inside the async `table()` wrapper after `emitTable` returns. The trailing blank line moved with it so the sync `dumpTables` path still emits it.

**#1, #2 (test coverage for check constraints / foreign keys):** out of scope for this PR — the goal is Rails-fidelity refactor of `schema_dumper.rb` private helpers, which mirrors Rails (`check_constraints_in_create`, `foreign_keys` are private helpers in `activerecord/lib/active_record/schema_dumper.rb` lines 283-346). The corresponding Rails tests live with the adapters that expose `check_constraints`/`foreign_keys` (Postgres/MySQL); unskipping our equivalents requires landing those adapter hooks first and is tracked separately.

deanmarano:
Copilot review #2:

**#1 (defineParams/formattedVersion unused) — fixed in 53adba4:** `dumpWithVersion` now threads the discovered version through dumper options, and `header()` emits `// version: <formatted>` via `defineParams()` when set. Both methods are now reachable on the live dump path.

**#2 (output stability) — PR description updated:** the new FK/check-constraint emission only fires when the adapter exposes `foreignKeys`/`checkConstraints` hooks. No current adapter does, so the existing on-disk dumps are byte-identical; the wiring is here so those hooks can land without another dumper PR. Body now calls this out explicitly.

**#3 (strip prefix/suffix from index name) — Rails source justification:** Rails does not strip prefix/suffix from `index.name`. See `activerecord/lib/active_record/schema_dumper.rb`:
```ruby
def index_parts(index)
  index_parts = [
    index.columns.inspect,
    "name: #{index.name.inspect}",
  ]
```
and `indexes_in_create` (lines 244-263) which calls `index_parts` directly with no name normalization. `remove_prefix_and_suffix` is only applied to `index.table` (in the standalone `indexes` method, line 235). Our implementation matches.

**#4 (strip prefix/suffix from FK name) — Rails source justification:** Same — Rails does not strip prefix/suffix from `foreign_key.name`:
```ruby
if foreign_key.export_name_on_schema_dump?
  parts << "name: #{foreign_key.name.inspect}"
end
```
(`schema_dumper.rb` lines 332-334). Rails filters out auto-generated `fk_rails_<hash>` names via `export_name_on_schema_dump?` (separate concern, tracked as follow-up alongside the `fk_ignore_pattern` cattr we don't yet honor).

deanmarano:
Copilot review #3:

**#1, #2 (invalid emitted syntax) — fixed in cbe0d7d:** `addCheckConstraint` and `addForeignKey` now wrap option fragments in a single `{ ... }` object, so emitted lines are valid TS:
```ts
await ctx.addCheckConstraint("users", "age >= 0", { name: "chk", validate: false });
await ctx.addForeignKey("orders", "users", { column: "user_id", onDelete: "cascade" });
```

**#5 (formatIndexParts unused) — fixed in cbe0d7d:** `indexParts` now consumes `formatIndexParts` for the `length`/`order`/`opclass` fragments, exactly like Rails' `index_parts` (`schema_dumper.rb` lines 271-273). Optional `lengths`/`orders`/`opclasses`/`where`/`using` fields surfaced on `IndexInfo` so adapters can populate them.

**#3, #4 (strip prefix/suffix from index/FK *names*) — Rails source justification (re-asserted):** Rails does **not** strip prefix/suffix from `index.name` or `foreign_key.name`. `schema_dumper.rb`:
```ruby
# index_parts (lines 265-281)
index_parts = [index.columns.inspect, "name: #{index.name.inspect}", ...]

# foreign_keys (lines 332-334)
if foreign_key.export_name_on_schema_dump?
  parts << "name: #{foreign_key.name.inspect}"
end
```
`remove_prefix_and_suffix` is only applied to *target tables* (`index.table`, `foreign_key.from_table`, `foreign_key.to_table`). The PR description already calls this out (“applied to table, index target, and FK target-table names per Rails — Rails does NOT strip from index/FK *names*”). No change here without new evidence.

deanmarano:
Copilot review #4:

**#1 (scalar lengths/orders/opclasses) — Rails source justification:** Rails' `format_index_parts` explicitly handles both Hash and non-Hash forms (`schema_dumper.rb` lines 358-364):
```ruby
def format_index_parts(options)
  if options.is_a?(Hash)
    "{ #{format_options(options)} }"
  else
    options.inspect
  end
end
```
The corresponding adapter type `PgIndexDefinition` already exposes `orders: string | Record<string, string>` (and Rails' `IndexDefinition` allows the scalar form for single-column indexes). Narrowing `IndexInfo` to `Record`-only also broke the existing `PostgreSQLAdapter`/`SchemaSource` compatibility (`tsc --build` failure) — keeping the union matches both Rails and the existing downstream contract.

**#2 (column-name keys may not be valid identifiers) — fixed in e7938e9:** `formatOptions` now JSON-quotes keys that don't match `/^[a-zA-Z_$][\w$]*$/`, so column names like `user-id` emit as `"user-id": 10`.

**#3 (deferrable dropped) — fixed in e7938e9:** `foreignKeys` now emits `deferrable: ...` when truthy, mirroring `schema_dumper.rb` line 338:
```ruby
parts << "deferrable: #{foreign_key.deferrable.inspect}" if foreign_key.deferrable
```

deanmarano:
Copilot review #5:

**#1 (hooks ignored on non-AdapterSchemaSource sources) — fixed in f7b4db9:** Centralized hook resolution in `_hookHost("foreignKeys" | "checkConstraints")`, which checks both the `SchemaSource` directly and the wrapped adapter (when the source is an `AdapterSchemaSource`). `SchemaStatements`-backed sources, which expose `foreignKeys`/`checkConstraints` on themselves, now contribute too.

**#2 (PR description claim that no adapter implements the hooks) — fixed:** The previous wording was wrong — Postgres, MySQL2, SQLite3, and `SchemaStatements` all expose these hooks. Updated the PR body to call out that the dump now emits `addForeignKey`/`addCheckConstraint` lines for tables that have FKs/check constraints (matching Rails). The existing schema-dumper tests don't seed FKs/check constraints so their snapshots are unchanged; the new emission is the intended Rails-fidelity behavior, not a hidden surprise.

deanmarano:
Copilot review #6:

**#1 (sync path skipped tableName / hook calls) — partially fixed in b9dc54f:** Sync `dumpTables` now sets/resets `this.tableName` around `emitTable` so subclass hooks see the current table. `checkConstraintsInCreate`/`foreignKeys` remain async-only — sync `SchemaSource` implementations (`MigrationContext`) don't expose those hooks, and turning a sync iteration into an async one to call them would propagate Promise returns through the public sync `dump()` overload. The async path (the one used by every adapter) already calls both, matching Rails.

**#2 (length/opclass/using not in MigrationContext.addIndex types) — Rails source justification:** `indexParts` mirrors Rails' `schema_dumper.rb#index_parts` (lines 265-281) which emits length/order/opclass/where/using/include/nulls_not_distinct/type/comment. The optional fields on `IndexInfo` are populated by SchemaSource implementations, but `AdapterSchemaSource.indexes()` currently forwards only `columns/unique/name` — so in the live adapter dump path these new options never fire today. Widening `MigrationContext.addIndex`'s option type so dumped schemas typecheck cleanly when adapters start populating those fields is the correct downstream change, but it's a separate scoped PR (touches the addIndex public surface, not the dumper).

deanmarano:
Copilot review #7 — these are restatements of review #6 #1 and #2. No new evidence; prior decisions stand:

**#1 (length/order/opclass/using vs MigrationContext.addIndex types):** `indexParts` mirrors Rails `schema_dumper.rb#index_parts` (lines 265-281). `AdapterSchemaSource.indexes()` currently only forwards `columns/unique/name`, so the extra fields don't fire in any live dump path. Widening `MigrationContext.addIndex`'s option type to match the full Rails surface is a separate, scoped change against `MigrationContext` — not the dumper. See [prior reply](https://github.com/blazetrailsdev/trails/pull/1111#issuecomment-4364489923).

**#2 (sync path skips checkConstraints/foreignKeys):** Sync `SchemaSource` implementations (`MigrationContext`) don't expose `checkConstraints`/`foreignKeys` hooks. Routing the sync path through an async `table()` would propagate Promise returns through `SchemaDumper.dump()`'s sync overload — that's a public-API change, not a fidelity gap. The sync `tableName` set/reset added in b9dc54f keeps the visible state consistent across both paths; hook emission is async-only by design because the hooks themselves are async. See [prior reply](https://github.com/blazetrailsdev/trails/pull/1111#issuecomment-4364489923).

PR #1112, opened 2026-05-02
feat(activerecord): PR A — Rails-named privates on Relation (assert_modifiable!, arel_column*)

deanmarano:
## Summary

First slice of the activerecord-privates push (see audit thread). Introduces 7 Rails-mirrored private helpers on `Relation` plus the `UnmodifiableRelation` error:

- `assertModifiableBang` — guard for mutation builders
- `checkIfMethodHasArgumentsBang` — empty-args guard for chainable scope-builders
- `isTableNameMatches` — `from()` clause table-name regex check
- `arelColumn` / `arelColumns` / `arelColumnsFromHash` / `arelColumnWithTable` — canonical column resolver shared (in Rails) by `select`/`group`/`pluck`/aggregates

Helpers are not yet wired into existing call sites — `_toSqlWithoutSetOp`, `groupColumnToArel`, etc. continue to do their own resolution. PR A2 will route those through the new helpers; this PR establishes the surface and tests so the api:compare needle moves and reviewers can scrutinize the shape independently.

activerecord privates: **627/1725 → 634/1725** (+7).

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test --project activerecord` — 9973 passed, 17 new tests in `build-arel-helpers.test.ts`
- [x] `pnpm api:compare --package activerecord --privates-only` — relation.rb 20→27 / 107

### DIFF
--- a/packages/activerecord/src/errors.ts
+++ b/packages/activerecord/src/errors.ts
@@ -495,6 +495,13 @@ export class TransactionIsolationError extends ActiveRecordError {
   }
 }
 
+export class UnmodifiableRelation extends ActiveRecordError {
+  constructor(message = "This relation is unmodifiable", options?: ErrorOptions) {
+    super(message, options);
+    this.name = "UnmodifiableRelation";
+  }
+}
+
 export class IrreversibleOrderError extends ActiveRecordError {
   constructor(message?: string, options?: ErrorOptions) {
     super(message, options);
--- a/packages/activerecord/src/index.ts
+++ b/packages/activerecord/src/index.ts
@@ -213,6 +213,7 @@ export {
   TransactionIsolationError,
   IrreversibleOrderError,
   UnknownAttributeReference,
+  UnmodifiableRelation,
 } from "./errors.js";
 export {
   ReadonlyAttributeError,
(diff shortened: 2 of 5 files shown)


deanmarano:
Re Copilot comment #1 on `isTableNameMatches`'s `(?<!FROM)\s` lookbehind:

This is intentional and matches Rails verbatim. Source — `activerecord/lib/active_record/relation/query_methods.rb:2009`:

```ruby
def table_name_matches?(from)
  table_name = Regexp.escape(table.name)
  quoted_table_name = Regexp.escape(model.adapter_class.quote_table_name(table.name))
  /(?:\A|(?<!FROM)\s)(?:\b#{table_name}\b|#{quoted_table_name})(?!\.)/i.match?(from.to_s)
end
```

The negative lookbehind is deliberate. `table_name_matches?` is asking 'does the model's table appear *somewhere other than as the from-target itself*?' — i.e. in a JOIN/WHERE inside a subquery passed to `from(...)`. If the only occurrence is the literal `FROM posts`, qualifying bare attribute names to `posts` would cause double-qualification against the user's chosen alias. Matching elsewhere (e.g. `SELECT … FROM (SELECT * FROM users WHERE users.id = 1) sub`) is the intended hit. Switching to a positive `FROM\s+` match would invert that semantics. No change.

PR #1115, opened 2026-05-02
fix(activemodel): pass defaultAttributes through LazyAttributeHash (P1)

deanmarano:
## Summary

Implements **P1** from `docs/activemodel-alignment.md` (audit-core §12 — highest-severity finding).

Previously `LazyAttributeHash` accepted only `(types, values)`, so on lazy materialize an attribute absent from `values` always became `Uninitialized` — silently dropping schema-level column defaults. This brings the constructor in line with Rails: `(types, values, additionalTypes, defaultAttributes, delegateHash)`.

Lookup precedence on lazy materialize now matches Rails `builder.rb:165-179`:
1. `values[name]` → `Attribute.fromDatabase`
2. `defaultAttributes[name]` → cloned (detached from prototype)
3. `Attribute.uninitialized(name, type)`

`Builder#buildFromDatabase` also gained an optional `additionalTypes` parameter so the eager path mirrors Rails' `LazyAttributeSet` precedence.

## Changes

- `packages/activemodel/src/attribute-set/builder.ts` — extend `LazyAttributeHash` constructor + `assignDefault`, thread `additionalTypes` through `Builder#buildFromDatabase`, update `marshalDump` / `marshalLoad` / `deepDup` for the new fields. Added `dupAttribute` helper (`@internal`).
- `packages/activemodel/src/attribute-set/builder-defaults.test.ts` — new tests covering the four required cases.

## Test plan

- [x] `pnpm --filter @blazetrails/activemodel test` — 1653/1653 pass
- [x] `pnpm api:compare --package activemodel` — **625/625 (100%)**, up from 621/625
- [x] `pnpm lint` clean
- [x] `pnpm format:check` clean

## Notes

LOC ~110 — under the 300-LOC ceiling. No `attribute-registration.ts` ripple was needed: `Builder` already accepts `defaultAttributes` in its constructor; the gap was that `LazyAttributeHash` ignored them on materialize.

### DIFF
--- a/packages/activemodel/src/attribute-set/builder-defaults.test.ts
+++ b/packages/activemodel/src/attribute-set/builder-defaults.test.ts
@@ -31,6 +31,39 @@ describe("LazyAttributeHash defaultAttributes", () => {
     expect(hash.get("age").isInitialized()).toBe(false);
   });
 
+  it("Builder#buildFromDatabase casts a present value using additionalTypes override", () => {
+    const types = new Map([["score", strType]]);
+    const builder = new Builder(types);
+    const additional = new Map([["score", intType]]);
+    const set = builder.buildFromDatabase({ score: "42" }, additional);
+    expect(set.fetchValue("score")).toBe(42);
+  });
+
+  it("LazyAttributeHash uses additionalTypes for present values", () => {
+    const hash = new LazyAttributeHash(
+      new Map([["score", strType]]),
+      { score: "42" },
+      new Map([["score", intType]]),
+    );
+    expect(hash.get("score").value).toBe(42);
+  });
+
+  it("marshalDump/marshalLoad round-trips all five fields", () => {
+    const types = new Map([
+      ["status", strType],
+      ["score", strType],
+    ]);
+    const additional = new Map([["score", intType]]);
+    const defaults = new Map([["status", Attribute.withCastValue("status", "active", strType)]]);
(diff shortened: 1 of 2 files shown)


deanmarano:
Copilot review #1 responses:

**C1 — types.has(name) check in assignDefault**: kept as-is. Rails `builder.rb:172` writes `Uninitialized` only when `types.key?(name)`. Keys present only in `additional_types` return nil from `assign_default_value` (no else branch on lines 165-179). The current trails behavior — `Attribute.null(name)` for unknown keys — is the closest TS equivalent and a pre-existing path; widening the check would diverge from Rails.

**C2 — deepDup shares additionalTypes/defaultAttributes by reference**: kept as-is. Rails `LazyAttributeHash#initialize_dup` (lines 124-127) only re-copies `@delegate_hash`; `@types`, `@values`, `@additional_types`, `@default_attributes` are shared by reference via Ruby's default `dup`. Trails matches this. (`LazyAttributeSet.deepDup` cloning `additionalTypes` is a separate, pre-existing divergence — out of P1 scope.)

**C3 — `name in values` → hasOwnProperty**: fixed in 76132e2.

**C4 — additionalTypes test coverage**: added tests for both eager (`Builder#buildFromDatabase`) and lazy (`LazyAttributeHash#get`) paths in 76132e2.

**C5 — marshal round-trip test**: added in 76132e2.

PR #1116, opened 2026-05-02
feat(arel): PR 32 — structural cleanup (Quoted/With/ValuesList/Exists/Function/Table/RollUp)

deanmarano:
## Summary

Structural alignment of several Arel node classes with their Rails counterparts. No new public API — these are inheritance and slot-storage changes that make the TS types match Rails' class hierarchy.

- **`Quoted extends Unary`** (`casted.rb:39`): value now stored in the `expr` slot; `get value()` shim preserves the public API.
- **`With` / `ValuesList` use `Unary#expr` slot**: children/rows stored in `expr` instead of a parallel property; getters expose them at the same name.
- **`Exists extends Function`** (`function.rb:47`): was a standalone `Node`; now a `Function` subclass with `expressions: Node[]`. `visitArelNodesExists` updated to visit `expressions[0]`.
- **`Function.alias=` wraps strings in `SqlLiteral`** (`function.rb:25`): converted to getter/setter; assigning a string auto-wraps.
- **`Table.eql()` / `Table.hash()`** (`table.rb:88–99`): compares name + tableAlias, hash is name-only (Rails excludes aliases to avoid loops).
- **`RollUp` is now the canonical class** (`roll_up.rb`): `Rollup` remains as a deprecated `const` alias. `postgresql.ts` visitor parameter type updated.

**Mechanical rename note:** `Rollup → RollUp` matches Rails casing. `Nodes.Rollup` still works via the alias.

## Test plan

- [ ] `pnpm vitest run --project other "packages/arel/src/"` — 1358 tests pass
- [ ] New `function.test.ts` covers: `alias=` setter, `RollUp` canonical name + `Rollup` alias, `With` expr slot, `Exists extends Function`
- [ ] `casted.test.ts` extended: `Quoted extends Unary`, value getter, eql
- [ ] `table.test.ts` extended: eql by name, eql by tableAlias, hash stability

### DIFF
--- a/packages/arel/src/nodes/casted.test.ts
+++ b/packages/arel/src/nodes/casted.test.ts
@@ -4,6 +4,24 @@ import { buildQuoted } from "./casted.js";
 import { Attribute as AMAttribute, ValueType } from "@blazetrails/activemodel";
 import { SelectManager } from "../select-manager.js";
 
+describe("Arel::Nodes::Quoted", () => {
+  it("is a Unary subclass (value stored in expr slot)", () => {
+    const q = new Nodes.Quoted(42);
+    expect(q).toBeInstanceOf(Nodes.Unary);
+  });
+
+  it("value getter returns expr", () => {
+    const q = new Nodes.Quoted("hello");
+    expect(q.value).toBe("hello");
+    expect((q as { expr: unknown }).expr).toBe("hello");
+  });
+
+  it("eql compares by value", () => {
+    expect(new Nodes.Quoted(1).eql(new Nodes.Quoted(1))).toBe(true);
+    expect(new Nodes.Quoted(1).eql(new Nodes.Quoted(2))).toBe(false);
+  });
+});
+
 describe("#hash", () => {
   const users = new Table("users");
   it("is equal when eql? returns true", () => {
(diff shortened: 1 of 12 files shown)


deanmarano (commented):
**Comment 1 — `Unary<TExpr>` generic vs `unknown`:**

A generic `Unary<TExpr>` would fragment the node hierarchy: every visitor method that accepts `Nodes.Unary` would need to become `Nodes.Unary<Node>` (or a union), and the dispatch cache keys on the runtime class — making generics transparent at runtime but painful in the type signatures. Rails itself stores `expr` as a plain Ruby object with no type constraint.

The widening added exactly **2 new casts** in visitors (`ordering.expr as Nodes.NodeOrValue` in `mysql.ts` and `node.expr as Nodes.NodeOrValue` in `to-sql.ts`) — both in methods that already knew the runtime type via the parameter type. Every other consumer that read `expr` as a `Node` already had `instanceof Node` guards (see `visitArelNodesOffset`, `visitArelNodesBin`, `visitArelNodesGrouping`, etc.). The practical blast radius is minimal and the code is cleaner than a pervasive generic.

**Comment 2 — `type Rollup = RollUp` backward compat:** Fixed — both a `const` (value namespace) and a `type` alias now exported.

deanmarano (commented):
Same suggestion as review #2 — already addressed there. To summarise: a generic `Unary<TExpr>` is impractical because the visitor dispatch is class-keyed at runtime and every visitor method would need type parameters or union-based overloads. Rails stores `expr` as a plain Ruby object with no type constraint. The widening to `unknown` required exactly 2 added casts in the codebase; all other `expr` consumers already guarded with `instanceof Node` or explicit casts. No further change needed.

PR #1117, opened 2026-05-02
fix(activemodel): preserve attribute state through withType/withValueFromUser (P2)

deanmarano:
## Summary

P2 of the activemodel attribute-lifecycle alignment chain. Fixes audit findings #3, #4, #5 from `docs/activemodel/audit-core.md` §8 — three bugs in `attribute.ts` that silently corrupted attribute state. See `docs/activemodel-alignment.md` P2 for the full spec.

- **`withType` drops in-place changes / originalAttribute** → mirror Rails `attribute.rb:91-97`: branch on `changedInPlace()`, route through `withValueFromUser(value).withType(type)` when mutated; otherwise reconstruct via `this.constructor` preserving `originalAttribute`. Previously always re-built via `Attribute.withCastValue`, losing both.
- **`withValueFromUser` skips `assertValidValue`** → mirror Rails `attribute.rb:78-81`: call `this.type.assertValidValue(value)` before constructing the new attribute. Invalid values now throw at assignment time instead of at first read.
- **`Uninitialized#originalValue` returned `undefined`** → mirror Rails `attribute.rb:255-257`: return the new `UNINITIALIZED_ORIGINAL_VALUE` sentinel (`Symbol.for`-based singleton, exported from `attribute.ts` and re-exported from the package index). AR dirty tracking can now identity-check against it.
- Added `Uninitialized#withType` and `Null#withType` overrides to mirror Rails subclass-specific behavior (`attribute.rb:231-233`, `:270-272`).

## Test plan

- [x] `pnpm exec vitest run packages/activemodel` — 1661 / 1661 passing
- [x] `pnpm exec eslint packages/activemodel` — clean
- [x] `pnpm --filter @blazetrails/activemodel build` (typecheck) — clean
- [x] `pnpm api:compare --package activemodel` — 625 / 625 (100%, up from 621)
- [x] `pnpm format:check` — clean
- [x] New tests in `attribute.test.ts`: in-place mutation preserved through `withType`; `originalAttribute` preserved through `withType`; `withValueFromUser` rejects invalid values via `assertValidValue`; `Uninitialized#originalValue === UNINITIALIZED_ORIGINAL_VALUE` (identity); sentinel singleton across import paths.

### DIFF
--- a/packages/activemodel/src/attribute.test.ts
+++ b/packages/activemodel/src/attribute.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from "vitest";
 import { Model, Types } from "./index.js";
-import { Attribute } from "./attribute.js";
+import { Attribute, FromUser, UNINITIALIZED_ORIGINAL_VALUE } from "./attribute.js";
+import { UNINITIALIZED_ORIGINAL_VALUE as UNINITIALIZED_FROM_INDEX } from "./index.js";
 import { typeRegistry } from "./type/registry.js";
 import "./attribute/user-provided-default.js";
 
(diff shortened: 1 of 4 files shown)


deanmarano:
Re: Copilot review #3 — correct that the in-place branch is currently dormant in trails, but that's because `Attribute.changedInPlace()` is hardcoded to `false` in the base class (attribute.ts:90-92), not because `withType` is wrong. Rails wires this through the type: `def changed_in_place?; has_been_read? && type.changed_in_place?(original_value_for_database, value); end` (attribute.rb:70-72). Trails has the type-side helper (`type/helpers/mutable.ts`) but never plumbs `type.changedInPlace?(rawOldValue, newValue)` back into `Attribute#changedInPlace?`. That's a pre-existing divergence, separate from the lifecycle fix.

P2's `withType` faithfully mirrors Rails' branching (attribute.rb:91-97). Once `changedInPlace?` is wired up in a follow-up, the existing branch activates with no further changes here. Verified by the new `MutableFromUser` test: when `changedInPlace()` returns true, `withType` does route through `withValueFromUser(this.value).withType(type)` and the mutation is preserved.

PR #1118, opened 2026-05-02
fix(activemodel): return ValueType fallback for unknown attributeTypes lookups (P3)

deanmarano:
## Summary

Implements P3 from `docs/activemodel-alignment.md` (audit-core §7, finding #11).

`attributeTypes()` previously returned a plain `Record<string, Type>`, so unknown keys read as `undefined` — meaning `attributeTypes(name).cast(value)` blew up with `Cannot read properties of undefined` on any unregistered attribute. Rails sets `hash.default = Type.default_value` on the cast-types hash so unknown lookups return a `Value` instance instead.

We mirror Rails by wrapping the cast-types record in a `Proxy` whose `get` trap returns `typeRegistry.lookup("value")` for any string key that isn't in the underlying record. `typeForAttribute(name)` now consequently returns `Type` (no longer `Type | null`); the AR base override does the same with the same fallback.

Dead `?? new ValueType()` guard removed in `type-caster/map.ts` now that the lookup never returns nullish.

This also unblocks P24a's schema-drift policy (YAML/JSON envelopes can fall back to `value` when the registry doesn't know a type key).

## Changes

- `packages/activemodel/src/attribute-registration.ts` — `attributeTypes()` returns a Proxy with `Type.default_value` fallback; `typeForAttribute()` typed as `Type`.
- `packages/activemodel/src/model.ts` — instance `typeForAttribute` return type tightened.
- `packages/activerecord/src/base.ts` — static override returns ValueType fallback.
- `packages/activerecord/src/type-caster/map.ts` — drop dead `?? new ValueType()` guard.
- Tests updated and added in `attribute-registration.test.ts` / `attributes.test.ts`.

## Test plan

- [x] `pnpm --filter @blazetrails/activemodel test` (1651 tests pass)
- [x] `pnpm --filter @blazetrails/activerecord test` (9956 pass / 3292 skipped)
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm api:compare --package activemodel` — 625/625 (100%)
- [x] `pnpm format:check`

## References

- `docs/activemodel-alignment.md` — P3
- `docs/activemodel/audit-core.md` §7 (#11)
- Rails: `activemodel/lib/active_model/attribute_registration.rb:37-51`

### DIFF
--- a/packages/activemodel/src/attribute-registration.test.ts
+++ b/packages/activemodel/src/attribute-registration.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { Model, Types } from "./index.js";
+import { Model, Types, ValueType } from "./index.js";
 
 describe("AttributeRegistrationTest", () => {
   it("attributes can be registered", () => {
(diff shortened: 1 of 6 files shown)


deanmarano:
Re Copilot review #2 (Proxy `get` trap own-property check):

Keeping `Object.hasOwn` — this matches Rails Hash semantics, which is what we're mirroring. Rails sets `hash.default = Type.default_value` (`active_model/attribute_registration.rb:38-40`), and Ruby Hash#[] only considers own keys; the default fires for *every* missing key including names that happen to collide with Object prototype methods. Switching to `prop in target` would make `attributeTypes()["toString"]` return `Object.prototype.toString` instead of the ValueType fallback — the opposite of Rails behavior.

This also directly contradicts Copilot review #1 on the same line, which (correctly) flagged `prop in target` as the bug and asked for `Object.hasOwn`. No callers in the codebase call `.toString()` on the attribute-types record; lookups are always attribute names from the schema.

PR #1120, opened 2026-05-02
fix(arel): PR A — Case#then, BoundSqlLiteral#plus, Composite block forwarding

deanmarano:
## Summary
Three independent leaf-node fidelity fixes against Rails 8.0.2 Arel, sourced from a recent file-by-file audit:

- **`Case#then(result)`** — port Rails' `Case#then(expression)`. Sets the right side of the most recent `When` clause; throws if called before `#when`.
- **`BoundSqlLiteral#plus(other)`** — port Rails' `BoundSqlLiteral#+`, returning a `Fragments` node containing self and the other node. Method-renamed to `plus` because TS classes can't define an arithmetic operator (consistent with `SqlLiteral#plus` already in the codebase).
- **`Composite#addBind`** — forward the `block` argument to both child collectors (was only forwarding to `left`). Matches Rails behavior.

Each fix is localized to one file plus a focused test. 70 LOC, no visitor or AST shape changes.

## Test plan
- [x] New unit tests for each fix (6 cases)
- [x] `pnpm vitest run --project other packages/arel` — 1363/1363 passing
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean

### DIFF
--- a/packages/arel/src/collectors/composite.test.ts
+++ b/packages/arel/src/collectors/composite.test.ts
@@ -14,6 +14,22 @@ describe("TestComposite", () => {
     expect(binds.value).toEqual([123]);
   });
 
+  it("addBind forwards block to both collectors", () => {
+    const left = new Collectors.SQLString();
+    const calls: number[] = [];
+    const right = {
+      append: () => right,
+      addBind: (_v: unknown, block?: (i: number) => string) => {
+        if (block) calls.push(1);
+        return right;
+      },
+    };
+    const composite = new Collectors.Composite(left, right);
+    composite.addBind(42, (i) => `$${i}`);
+    expect(left.value).toBe("$1");
+    expect(calls).toEqual([1]);
+  });
+
   it("retryable on composite collector propagates", () => {
     const sql = new Collectors.SQLString();
     const binds = new Collectors.Bind();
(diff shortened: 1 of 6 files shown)


deanmarano:
Re: Copilot review #2 — keeping the `instanceof Node` guard.

Rails' `Arel.arel_node?` is precisely defined (`activerecord-8.0.2/lib/arel.rb:64-66`):

\`\`\`ruby
def self.arel_node?(value)
  value.is_a?(Arel::Nodes::Node) || value.is_a?(Arel::Attribute) || value.is_a?(Arel::Nodes::SqlLiteral)
end
\`\`\`

That set does NOT include `SelectManager` / `TreeManager`. In our TS port, `Attribute` and `SqlLiteral` both extend `Node`, so `instanceof Node` already accepts the full Rails-defined arel-node set (and is in fact slightly broader). Accepting `{ ast: Node }` shapes here would *exceed* Rails' contract for `BoundSqlLiteral#+`.

Rails would raise `ArgumentError("Expected Arel node")` if a `SelectManager` were passed to `BoundSqlLiteral#+`. Our TS version does the same, by design.

The `buildQuoted` unwrap of `.ast` happens at a different boundary (predicate construction, where Rails relies on Ruby's duck-typing through `is_a?` chains in `Predications`). It is not part of `arel_node?`.

No change.

PR #1153, opened 2026-05-04
fix(activemodel): truncate odd/even via Math.trunc + gate raw value on cameFromUser (P16)

deanmarano:
Closes audit findings #28 and #29 from `docs/activemodel/audit-validations.md` §15. See `docs/activemodel-alignment.md` P16.

## Changes

**`packages/activemodel/src/validations/numericality.ts`**
- `odd`/`even` now truncate via `Math.trunc` before modulo check, matching Rails `value.to_i.odd?` / `value.to_i.even?` (numericality.rb:51). `Math.trunc(-2.5) === -2` matches Ruby `-2.5.to_i == -2`; `Math.floor` would give -3 (wrong).
- `prepareValueForValidation` now mirrors Rails numericality.rb:120-141: if host exposes `cameFromUser(name)`, branches on it to use `readAttributeBeforeTypeCast` (user input) or `readAttribute` (cast DB value). Hosts without `cameFromUser` fall back to the previous `readAttributeBeforeTypeCast` path — non-AR hosts degrade gracefully.
- `RecordWithRawAttribute` interface gains optional `cameFromUser?(name: string): boolean`.

**`packages/activerecord/src/attribute-methods/before-type-cast.ts`**
- `isAttributeCameFromUser` exported (was unexported private).

**`packages/activerecord/src/base.ts`**
- Imports `isAttributeCameFromUser` and assigns it as `cameFromUser` instance method on Base, wiring the AR AttributeSet's per-attribute `cameFromUser()` flag into the validator dispatch chain.

## Impact

- `2.5` with `even: true` now passes (was incorrectly failing — `2.5 % 2 !== 0` without truncation).
- Records loaded from DB validate against the cast number, not the raw column string.
- 7 new tests; 58 → 65 in numericality-validation.test.ts.
- LOC delta: 127+8 = 135, well within 300-LOC budget.

### DIFF
--- a/packages/activemodel/src/validations/numericality-validation.test.ts
+++ b/packages/activemodel/src/validations/numericality-validation.test.ts
@@ -1,5 +1,6 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import { Model } from "../index.js";
+import { prepareValueForValidation } from "./numericality.js";
 
 describe("NumericalityValidationTest", () => {
   it("validates numericality with greater than or equal using string value", () => {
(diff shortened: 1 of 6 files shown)


deanmarano:
Copilot comment 1 (**alias resolution in `isAttributeCameFromUser`**): Fixed in e9d639e. Added `this.constructor._attributeAliases?.[attrName] ?? attrName` resolution before calling `getAttribute`, consistent with how `readAttributeForDatabase` already handles aliases in the same file. Rails' private dispatch target `attribute_came_from_user?` (before_type_cast.rb:101) receives the already-resolved attribute name — the generated `nickname_came_from_user?` method wires through to `attribute_came_from_user?(:name)` internally. Trails calls `cameFromUser` directly with whatever the validator passes in, so alias resolution at the public boundary is the right level.

PR #1154, opened 2026-05-04
feat(activerecord): abstract/database-statements perform_query protocol (PR 4)

deanmarano:
## Summary

Implements the 8 private query-execution-core methods from `abstract/database_statements.rb` (lines ~555–745 of Rails 7.2 source), promoting non-exported stubs to real `export function` declarations so `api:compare` counts them.

**8 methods implemented:**
- `performQuery` — abstract hook; raises `NotImplementedError` per Rails contract; adapter subclasses override
- `preprocessQuery` — calls `checkIfWriteQuery` + `markTransactionWrittenIfWrite`; no-op if host skips them
- `select` — delegates to `internalExecQuery` (async `FutureResult` path deferred)
- `sqlForInsert` — appends `RETURNING` clause when `supportsInsertReturning()`, extracts table ref when `pk` is null
- `extractTableRefFromInsertSql` — regex table-name extraction from INSERT SQL
- `defaultInsertValue` — returns `Arel.sql("DEFAULT")` literal
- `returningColumnValues` — returns `[lastInsertedId(result)]`
- `arelFromRelation` — duck-types `.arel()` on Relation-like objects; pass-through otherwise

**`DatabaseStatementsHost` additions:** `checkIfWriteQuery`, `supportsInsertReturning`, `quoteColumnName`, `primaryKey`, `preprocessQuery`

**Deferred to PR 4b:** `buildFixtureSql`, `buildFixtureStatements`, `buildTruncateStatement`, `buildTruncateStatements`, `combineMultiStatements`

## performQuery signature

```ts
export function performQuery(
  this: DatabaseStatementsHost,
  _rawConnection: unknown,
  _sql: string,
  _binds: unknown[],
  _typeCastedBinds: unknown[],
  _options?: { prepare?: boolean; notificationPayload?: unknown; batch?: boolean },
): never
```

Adapter subclasses (PRs 17/18/18b) can safely super-delegate into this base. The only deviation from Rails is collapsing the three keyword args (`prepare:`, `notification_payload:`, `batch:`) into a single options object — standard TS idiom, no semantic change.

## Metrics

| | Before | After |
|---|---|---|
| Matched | 3810 (75.0%) | 3862 (76.0%) |
| `abstract/database-statements.ts` | 59/72 (82%) | 67/72 (93%) |
| Missing (this file) | 13 | 5 |
| LOC | — | +246/-51 = 297 total |
| Tests | 10015 passed | 10015 passed |

No files regressed from 100%.

## Misplaced moves

`api:compare` shows 0 misplaced after M2b (#1152). The `initialize` and `resetTransaction` moves noted in the plan were already resolved by prior PRs — no moves needed in this PR.

Refs `docs/activerecord-100-plan.md` Wave 2 PR 4

### DIFF
--- a/packages/activerecord/src/connection-adapters/abstract/database-statements.test.ts
+++ b/packages/activerecord/src/connection-adapters/abstract/database-statements.test.ts
@@ -504,11 +504,14 @@ describe("defaultInsertValue", () => {
 });
 
 describe("returningColumnValues", () => {
-  it("returns array with lastInsertedId result", () => {
-    const host: DatabaseStatementsHost = {
-      lastInsertedId: () => 42,
-    };
-    const fakeResult = Result.empty();
-    expect(returningColumnValues.call(host, fakeResult)).toEqual([42]);
+  it("returns [first value of first row] from result", () => {
+    const host: DatabaseStatementsHost = {};
+    const result = new Result(["id"], [[42]]);
+    expect(returningColumnValues.call(host, result)).toEqual([42]);
+  });
+
+  it("returns [undefined] for empty result", () => {
+    const host: DatabaseStatementsHost = {};
+    expect(returningColumnValues.call(host, Result.empty())).toEqual([undefined]);
   });
 });
--- a/packages/activerecord/src/connection-adapters/abstract/database-statements.test.ts
+++ b/packages/activerecord/src/connection-adapters/abstract/database-statements.test.ts
@@ -24,8 +24,17 @@ import {
   highPrecisionCurrentTimestamp,
   markTransactionWrittenIfWrite,
   isTransactionOpen,
+  performQuery,
+  preprocessQuery,
+  select,
+  sqlForInsert,
+  arelFromRelation,
+  extractTableRefFromInsertSql,
+  defaultInsertValue,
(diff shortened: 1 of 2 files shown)


deanmarano:
**Copilot review response (comments 1 & 2 — `returningColumnValues`):**

Both comments address the same point. The implementation is correct per Rails source.

`abstract/database_statements.rb` lines 719–725:
```ruby
def last_inserted_id(result)
  single_value_from_rows(result.rows)   # => first cell of first row, or nil
end

def returning_column_values(result)
  [last_inserted_id(result)]            # one-element array wrapping the scalar
end
```

The single-scalar-in-array shape is intentional for **non-RETURNING adapters** (MySQL, SQLite). The caller in `insert()` (line 199) unpacks this for the last-insert-ID path. PostgreSQL **overrides** `returning_column_values` with `result.rows.first` (its `database_statements.rb` line 208) to return all RETURNING columns — that override lands in PR 18b, not here.

`[undefined]` for empty results is correct: Rails returns `[nil]` when `single_value_from_rows([]) => nil`.

No changes needed. Implementation and tests are faithful to the base-class contract.

PR #1192, opened 2026-05-04
test(activerecord): harden reset to drop across all schemas/views (PR 2/4)

deanmarano:
## Summary

Addresses three legitimate gaps Copilot raised on PR 1's reset (#1190 review #2). All three were paths where state could survive `resetTestAdapterState()` and bleed into the next test:

1. **PostgreSQL** — was enumerating only `schemaname='public'`. Tests that create custom schemas (e.g. `schema.test.ts` with `test_schema`/`test_schema2`) leaked tables. Now enumerates every schema in `current_schemas(false)` (matches `PostgreSQLAdapter.tables()` scope) and drops with fully-qualified `"schema"."table"` identifiers.

2. **MySQL** — was iterating `SHOW TABLES` and calling `_sharedAdapter.exec()` per table. Each `exec()` may pick a different pool connection, so `SET FOREIGN_KEY_CHECKS=0` couldn't bracket the drops. Now pins a single dedicated pool connection for the whole reset, sets the FK flag once, drops views before tables, and releases the connection (which implicitly resets the session flag).

3. **SQLite** — was still using `_createdTables` + `dropAllTables()`. Tables created via raw `adapter.exec()` bypass the in-memory tracking set and survived reset. Now queries `sqlite_master` (excluding internal `sqlite_*`).

## Test plan
- [x] Full AR SQLite suite: 10203 passed / 3290 skipped, 0 regressions
- [ ] PostgreSQL Tests pass in CI
- [ ] MariaDB Tests pass in CI

## Plan context

PR 2 of the 4-PR series following #1190. Remaining: PR 3 kills the `handleMissingSchemaError` regex-based recovery path; PR 4 moves to schema-per-test isolation. After PR 4 stabilizes, the retry config from #1187 can be reverted.

### DIFF
--- a/packages/activerecord/src/test-adapter.ts
+++ b/packages/activerecord/src/test-adapter.ts
@@ -362,17 +362,9 @@ async function dropAllTables(inner: any): Promise<void> {
 }
 
 /**
- * Drop every test-created object in the named PG schema (matviews,
- * views, tables, standalone sequences, enums/domains/range types)
- * using the adapter's own identifier quoting. Used for `public`,
- * which the AR adapter assumes exists and so can't be dropped
- * wholesale via `DROP SCHEMA … CASCADE`.
- *
- * Order matters: matviews → views → tables (CASCADE handles their
- * cross-references). Then standalone sequences (those owned by a
- * column drop with the table) and finally types — enums/domains can
- * be referenced by table columns, so they must come after the table
- * drops.
+ * Drop every materialized view, view, and table in the named PG schema
+ * using the adapter's own identifier quoting (which doubles embedded
+ * quotes). Used for `public`, which the AR adapter assumes exists.
  *
  * @internal
  */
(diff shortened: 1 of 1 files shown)


deanmarano:
Reverted the last two commits (`b8c2a025c` switching to pg_namespace + DROP SCHEMA, `9070b738a` adding sequences/types) after CI run https://github.com/blazetrailsdev/trails/actions/runs/25350075916/job/74327603667 showed widespread test failures (`expected null`, `expected 1 to be 2`, `expected []`) on PG.

Those commits responded to Copilot's defensive concerns (search_path narrowness, schemas leaking, sequences/enums surviving) but I have no demonstrated test that exercises any of them — every failing test in this codebase uses `public`, and the only ATTACH-using test bypasses the shared adapter. The aggressive cleanup proved to break real tests instead of fixing hypothetical ones.

Keeping the safe first commit (`b748ef089`): drop matviews/views before tables, drop SQLite views, restore MySQL FOREIGN_KEY_CHECKS in finally with destroy-on-error fallback. Those are real correctness fixes, narrow in scope.

If/when a test starts exercising custom schemas or sequences/enums and bleeds state, we can extend reset then with a proven repro.

PR #1193, opened 2026-05-05
test(activerecord): instrument schema-recovery firings (PR 3/4)

deanmarano:
## Summary

Adds the scaffolding to kill the regex-based `handleMissingSchemaError` recovery path. **Direct removal isn't safe**: 45 firings across the SQLite suite today — mostly HABTM join tables (auto-generated, never registered as Models), STI subclasses, and a few seedPosts patterns. Each is a real test path that would break without recovery and needs its own follow-up fix.

## What ships

- Recovery-fire counter `_schemaRecoveryFires`, exported via `getRecoveryFireCount()` / `resetRecoveryFireCount()`.
- **Soft mode** (default): one-line `console.warn` diagnostic on every recovery firing — CI logs are now grep-friendly:
  ```
  [test-adapter] schema recovery fired: no such table: parrots_pirates | sql=INSERT INTO ...
  ```
- **Strict mode** (`AR_TEST_STRICT_SCHEMA=1`): throw with a message that names the SQL and tells the author exactly how to fix it (declare the column on the model via `attribute(...)`).

## What's next

Future PRs add explicit registration for the offending test-only patterns (HABTM join tables, STI subclasses, abstract-base shadow tables) and shrink the diagnostic count toward 0. Once strict CI is clean, the recovery path can be deleted in a PR-4-followup, and the retry config from #1187 can be reverted.

## Test plan
- [x] Full AR SQLite suite (soft mode): 10203 passed / 3290 skipped
- [x] Strict mode on `finder.test.ts` (no recovery): passes
- [x] Strict mode on `relations.test.ts` (recovery fires): correctly throws and names the SQL
- [ ] PG/MySQL CI: soft mode, no regressions

## Plan context

PR 3 of the 4-PR series following #1190 / #1192. Defers the recovery removal until the underlying registration gaps are closed.

### DIFF
--- a/packages/activerecord/src/test-adapter.ts
+++ b/packages/activerecord/src/test-adapter.ts
@@ -42,8 +42,13 @@ const isMysql = (): boolean => !!MYSQL_TEST_URL;
 
 /**
  * Tracks how many times the regex-based "missing table/column" recovery in
- * `handleMissingSchemaError` fired and patched a real schema mismatch. The
- * recovery path is the historical cause of cross-test drift (PRs #1190,
+ * `handleMissingSchemaError` *would have fired* (i.e. matched a recoverable
+ * error). In soft mode this also reflects DDL that ran; in strict mode the
+ * counter increments but no DDL runs (we throw first). The counter is about
+ * "schema mismatches the recovery path detected", not "schema patches
+ * applied" — both modes count the same event, just with different actions.
+ *
+ * The recovery path is the historical cause of cross-test drift (PRs #1190,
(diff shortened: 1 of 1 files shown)


deanmarano:
Superseded by the empirical result on #1194 (HABTM fix zeros recovery firings on SQLite — 45 → 0). The instrumentation was justified by the assumption we'd need to measure recovery before safely deleting it; the measurement turned out to be unnecessary. Skipping straight to PR 4e: delete handleMissingSchemaError + revert the retry config from #1187, after #1194 merges and PG/MySQL CI confirms no regressions.

PR #1194, opened 2026-05-05
test(activerecord): register HABTM join models with the test adapter (PR 4a/4)

deanmarano:
## Summary

Closes the HABTM gap that was causing every schema-recovery firing in the test-adapter. After this PR, the SQLite AR suite goes from **45 firings → 0**, measured against the instrumentation in #1193.

## Root cause

\`createHabtmJoinModel\` builds a JoinModel whose \`adapter\` property is a **getter** delegating to the LHS model:

\`\`\`ts
Object.defineProperty(JoinModel, "adapter", {
  get() { return lhsModel.adapter; },
  set(_v) { /* no-op */ },
  configurable: true,
});
\`\`\`

The Base.adapter setter is what normally fires \`_onAdapterSet\` (the test-adapter's registration hook). With a no-op setter the hook never fires for the JoinModel, so its join table never gets registered in \`_pendingModels\`, never gets a \`processPendingModels\` CREATE, and the first INSERT into it triggers the regex recovery path.

## Fix

- Add \`_fireAdapterSetHook(modelClass)\` to base.ts, the manual counterpart of the implicit setter-fired hook.
- \`createHabtmJoinModel\` calls it after attribute setup so the JoinModel goes through the same registration path as user-declared Models.

## Why this single change zeroes the counter

Cascade effects: many \`posts\` / \`children\` / \`dogs\` recovery firings I'd categorized as STI or seedPosts issues turned out to be downstream of HABTM cascades — fix-it-once, the rest follow.

## Test plan
- [x] Full AR SQLite suite: 10203 passed / 3290 skipped, recovery counter = 0
- [ ] PG/MySQL CI: still at 0 firings
- [ ] PR #1193 strict-mode CI job (once added) passes for SQLite

## Plan context

PR 4a of the 4-PR series following #1190 / #1192 / #1193. Originally I planned 4a-d covering HABTM, STI children, seedPosts-style re-registration, and a long tail. The empirical result (45 → 0 from HABTM alone) likely collapses 4a–d into just this PR. Once PG/MySQL CI confirm 0 firings, PR 4e can delete \`_handleMissingSchemaErrorInner\` outright and revert the retry config from #1187.

### DIFF
--- a/packages/activerecord/src/_adapter-set-hook.ts
+++ b/packages/activerecord/src/_adapter-set-hook.ts
@@ -1,0 +1,30 @@
+/**
+ * @internal
+ *
+ * Standalone module for the adapter-set hook so both base.ts and
+ * associations.ts can reach it without creating a circular value
+ * import. base.ts imports * as _Associations from associations.ts
+ * eagerly to wire .belongsTo/.hasOne/.hasMany; if associations.ts
+ * imports a value from base.ts, that value isn't defined yet on the
+ * first pass and the test "registers CollectionProxy ... after
+ * module reset" trips on `_Associations.belongsTo` being undefined.
+ *
+ * The hook itself is consumed by the test-adapter
+ * (packages/activerecord/src/test-adapter.ts), which calls
+ * `setOnAdapterSetHook(registerModel)` at module load. When a model's
+ * adapter is assigned, base.ts calls `fireAdapterSetHook(model)` to
+ * notify the test-adapter. HABTM's anonymous JoinModel uses a getter
+ * for `adapter`, so its setter is a no-op; createHabtmJoinModel
+ * fires the hook manually so the test-adapter learns about the
+ * join table just like any other model.
+ */
+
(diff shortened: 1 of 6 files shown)


deanmarano:
Reverted the globalThis pin commit (`955ed81d8`) — re-running 5x consecutive cross-file PG runs (finder + relations) on this branch shows 25/33/38/28/51 failures, not the 5/8/13/10/5 I claimed earlier. The earlier numbers were CI-variance noise, not real evidence the pin helped. I confirmed this independently by branching off main, applying the pin in isolation, and getting 32-44 failures on the same workload (worktree `test-adapter-pin`, not pushed).

The cross-file PG flakiness exists on main and on every active PR (verified: same failure shape on PR #1199's CI). My HABTM fix here doesn't make it worse on SQLite (still 10244 passed, 0 regressions) — the PG instability is orthogonal to this PR's diff and the right fix is a separate, larger investigation (likely schema-per-test isolation per the original 4-PR plan).

PR #1205, opened 2026-05-05
test(activerecord): remove schema-recovery and retry safety net (PR 4e)

deanmarano:
## Summary

Final follow-up to the flaky-test cleanup series (#1187 #1190 #1192 #1194 #1200). With eager HABTM JoinModel registration in place, the test-adapter's regex-based \`handleMissingSchemaError\` recovery path no longer fires (45→0 on SQLite; main is green on PG/MariaDB).

## Removes

- \`SchemaAdapter.handleMissingSchemaError\` (~120 lines): regex-driven CREATE TABLE / ADD COLUMN recovery from missing-relation/missing-column DB errors.
- \`SchemaAdapter.extractTableFromSql\` helper (only used by recovery).
- 5-attempt savepoint retry loops in \`execute\` / \`executeMutation\`.
- vitest \`retry: 2\` for PG/MySQL from #1187.

## Status: BLOCKED on eager-registration follow-ups

First CI confirmed the recovery path was masking real schema gaps. ~25 tests across ~13 files fail with "no such table" for tables the recovery path was silently materializing.

### Failure inventory (SQLite, first run)

Missing tables: \`children\`, \`dogs\`, \`orders\`, \`people\`, \`posts\`, \`projects\`, \`sub_sub_pirates\`, \`widgets\`.

Failing files:
- \`associations.test.ts\` (1) — habtm empty join
- \`base.test.ts\` (1) — composite-PK find on \`orders\`
- \`callbacks.test.ts\` (1) — STI inheritance (\`dogs\`/\`sub_sub_pirates\`)
- \`inherited.test.ts\` (2) — super before/after filter
- \`invertible-migration.test.ts\` (1) — change/down (\`widgets\`)
- \`migration.test.ts\` (2) — reversible \`renameTable\` (\`people\`)
- \`nested-attributes.test.ts\` (1) — overridden in subclasses
- \`relations.test.ts\` (10) — pluck subquery, find list of AR, eager-load size, etc.
- \`scoping/default-scoping.test.ts\` (1) — STI default scope
- \`scoping/named-scoping.test.ts\` (1) — subclass scope merge
- \`serialized-attribute.test.ts\` (1)
- \`signed-id.test.ts\` (2) — STI signed records

### Categories (root causes)

1. **STI subclasses** not in \`_registeredModelClasses\` when adapter sets — parent gets the hook, subclass doesn't share registration. Files: callbacks, inherited, default-scoping, named-scoping, signed-id.
[...]

PR #1207, opened 2026-05-05
test(activerecord): register posts table in last RelationTest block beforeEach

deanmarano:
## Summary

- The `describe("RelationTest")` block at line 4222 of `relations.test.ts` sets `adapter = freshAdapter()` in `beforeEach` but never wires it to the module-level `Post` class.
- Tests that call `Post.create()` directly (rather than defining a local class with `this.adapter = adapter`) found no `posts` table after `resetTestAdapterState()` dropped it.
- Fix: add `Post.adapter = adapter` to that block's `beforeEach`, which triggers `registerModel(Post)` so `setup()` creates the table eagerly — matching the pattern used by `seedPosts()` in the earlier describe blocks.

Unblocks #1205 by registering `relations.test.ts` fixture tables eagerly (10 tests that were silently recovered by `handleMissingSchemaError` now pass without the recovery path).

## Test plan

- [x] `pnpm vitest run packages/activerecord/src/relations.test.ts` — 675 passed, 9 skipped with cherry-pick of #1205's commit

### DIFF
--- a/packages/activerecord/src/relations.test.ts
+++ b/packages/activerecord/src/relations.test.ts
@@ -4223,6 +4223,7 @@ describe("RelationTest", () => {
   let adapter: DatabaseAdapter;
   beforeEach(() => {
     adapter = freshAdapter();
+    Post.adapter = adapter;
   });
 
   function makePost() {


deanmarano:

PR #1208, opened 2026-05-05
fix(activerecord): eager adapter registration for subclasses

deanmarano:
## Summary

- **`base.ts` infra fix**: the `Base.adapter` setter used `this._adapter === adapter` to skip no-op reassignments, but `this._adapter` resolves via the prototype chain. A subclass calling `this.adapter = parent.adapter` silently matched the inherited reference and returned early — the hook never fired and the subclass table was never registered for test-schema creation. Fixed with `hasOwnProperty` so the guard only applies to own-property assignments.
- **Test fixes**: 7 test files updated to remove reliance on the schema-recovery path being removed in #1205:
  - `inherited.test.ts`: Child classes set `this.adapter = adapter` so their tables are registered
  - `callbacks.test.ts`: Dog subclass sets `this.adapter = adapter`
  - `signed-id.test.ts`: STI tests use `enableSti` + `registerSubclass` (Dog/Car share parent table)
  - `scoping/default-scoping.test.ts`, `scoping/named-scoping.test.ts`: Dog subclasses set adapter
  - `serialized-attribute.test.ts`: VeryImportantTopic sets adapter; `nested-attributes.test.ts`'s SubSubPirate was already calling `this.adapter = adapter` but was silently no-op'd by the old guard — fixed by the infra change

Fixes the following failures surfaced by #1205 on SQLite:
`callbacks.test.ts` (dogs), `inherited.test.ts` (children), `signed-id.test.ts` (dogs, cars), `scoping/default-scoping.test.ts` (dogs), `scoping/named-scoping.test.ts` (dogs), `serialized-attribute.test.ts` (very_important_topics), `nested-attributes.test.ts` (sub_sub_pirates)

Companion PR: #1205 (do not merge before this)

## Test plan

- [ ] Cherry-pick #1205 onto this branch and run the 7 affected test files — all pass
- [ ] Existing tests unaffected (build + typecheck pass)

### DIFF
--- a/packages/activerecord/src/base.ts
+++ b/packages/activerecord/src/base.ts
@@ -733,9 +733,12 @@ export class Base extends Model {
    * infrastructure. Prefer `establishConnection` for production use.
    */
   static set adapter(adapter: DatabaseAdapter) {
-    // Reassigning the same adapter is a no-op — avoid dropping reflected
-    // columns/types unnecessarily when user code re-sets the same ref.
-    if (this._adapter === adapter) {
+    // Reassigning the same adapter to the same class is a no-op — avoid
+    // dropping reflected columns/types unnecessarily. Use hasOwnProperty so a
+    // subclass that calls `this.adapter = parent.adapter` still fires the hook
+    // and gets registered for table creation, even when the adapter reference
+    // matches what it inherited from the parent class.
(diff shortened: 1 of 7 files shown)


deanmarano:

PR #1209, opened 2026-05-05
fix(activerecord): eager table creation for migrations and ad-hoc fixtures

deanmarano:
## Summary

Removes reliance on the schema-recovery path (removed in #1205) in five test files:

- **`relations.test.ts`**: large `RelationTest` describe block created a fresh adapter each test but never re-assigned `Post.adapter`, leaving the module-level `Post` pointing at a cleaned-up adapter with no tables; add the assignment to `beforeEach`
- **`associations.test.ts`**: `test_habtm_returns_empty_when_no_join_rows` queries the "projects" table but never defined a `Project` model; add a minimal `Project` class so the table is created eagerly
- **`base.test.ts`**: `Order` CPK model declares `primaryKey = ["shop_id","id"]` but omits the `"id"` attribute, causing the `CREATE TABLE` to fail (column in `PRIMARY KEY` but not in column list); add the explicit attribute
- **`migration.test.ts`**: "reversible renameTable" tests renaming a "people" table that was never created; add setup `createTable`. "auto-reverses" test checked `SELECT * FROM widgets` after the migration dropped the table (relied on recovery returning empty rows); assert table absence instead
- **`invertible-migration.test.ts`**: same SELECT-after-drop pattern; assert table absence via adapter's table tracker

Fixes the following failures surfaced by #1205 on SQLite:
`relations.test.ts` (10 × posts), `associations.test.ts` (projects), `base.test.ts` (orders), `migration.test.ts` (people, widgets), `invertible-migration.test.ts` (posts)

Companion PR: #1205 (do not merge before this)

## Test plan

- [ ] Cherry-pick #1205 onto this branch and run the 5 affected test files — all pass
- [ ] Existing tests unaffected (build + typecheck pass)

### DIFF
--- a/packages/activerecord/src/associations.test.ts
+++ b/packages/activerecord/src/associations.test.ts
@@ -1484,7 +1484,14 @@ describe("HasAndBelongsToManyAssociations", () => {
         this.adapter = adapter;
       }
     }
+    class Project extends Base {
+      static {
+        this.attribute("name", "string");
+        this.adapter = adapter;
+      }
+    }
     registerModel(Developer);
+    registerModel(Project);
 
     // Create the join table
     await adapter.executeMutation(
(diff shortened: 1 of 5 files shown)


deanmarano:

PR #1220, opened 2026-05-05
fix(test): swap vi.stubEnv for per-file env save/restore in join-model

deanmarano:
## Status: investigation ongoing — do NOT merge

The vi.stubEnv hypothesis (that `vi.unstubAllEnvs` blast-radius across files in a worker was clearing the flag) **did not hold up under local PG repro**.

### What I confirmed locally on PG

Ran `PG_TEST_URL=... pnpm vitest run --project activerecord <join-model> <dirty>` (with this PR's `process.env.AR_NO_AUTO_SCHEMA = "1"` swap applied):

- 49 failures in join-model.test.ts: `relation "posts" does not exist`, `relation "authors" does not exist`, `column "author_id" of relation "posts" does not exist`, `duplicate key value violates unique constraint "pg_type_typname_nsp_index"`.
- A `console.log` inside `noAutoSchema()` confirmed the flag IS `"1"` during the failing tests — so the env swap is engaged correctly. The recovery path is genuinely off.
- Setting `fileParallelism: false` did NOT resolve it.

### What this tells us

- The failure is **not** caused by env-flag flipping.
- The failure happens even when noAutoSchema is correctly true throughout the failing test.
- Some tests in the file pass and others fail with similar shape — the failing ones are early in the file, later identical tests pass.
- The `pg_type_typname_nsp_index` duplicate-key errors point to **concurrent CREATE TABLE on the shared PG** — multiple sources racing on the same DDL.

### Why this PR's change isn't worthless

The `process.env.X` swap is still strictly safer than `vi.stubEnv` + `vi.unstubAllEnvs()` because the latter IS worker-global. But it doesn't fix the join-model PG failures — those have a different root cause that needs more digging (likely something in `_sharedAdapter` / `_createdTables` / `_pendingModels` shared module state under `AR_DB_FORKS=1`).

### Next steps

- Self-grade: **D** (hypothesis wrong, fix doesn't resolve cited issue).
- Leaving as draft. Will keep investigating the actual cause before reusing this PR or opening a new one.

### DIFF
--- a/packages/activerecord/src/associations/join-model.test.ts
+++ b/packages/activerecord/src/associations/join-model.test.ts
@@ -45,15 +45,14 @@ async function registerSchemaFor(adapter: DatabaseAdapter, ...models: any[]): Pr
 }
 
 // Disable the dynamic-adapter auto-schema path for this entire file. Must be
-// set before any describe/beforeEach runs so that createTestAdapter() and all
-// SchemaAdapter.setup() calls see the flag immediately.
+// set before any describe/beforeEach runs so createTestAdapter() and every
+// SchemaAdapter.setup() call sees the flag from the first test onward.
 //
-// We mutate process.env directly instead of vi.stubEnv because
-// vi.unstubAllEnvs() is global to the worker — when sibling test files in the
-// same fork (e.g. dirty/store/serialized-attribute after their TS-4 migration)
-// also call unstubAllEnvs in their own afterAll, they wipe this file's stub
-// too, racing this file's tests against an unset flag and producing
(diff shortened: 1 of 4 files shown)


deanmarano:
Closing — the vi.stubEnv hypothesis was wrong. Local PG repro shows the failure persists with this fix applied. MariaDB and SQLite pass; PG-specific issue remains. Reopening investigation under a new branch with targeted instrumentation; will open a fresh PR once root cause is identified.

deanmarano:
Reopening to run CI on the debug-instrumented branch. This is investigation-only — DO NOT MERGE. Will close again once root cause is identified and a real fix lands in a separate PR.

PR #1666, opened 2026-05-16
feat(activerecord): HABTM Slot F — eager loading (includes/preload/eager_load)

deanmarano:
## Summary

Wires `hasAndBelongsToMany` into `JoinDependency` so `.eagerLoad()` (and `.includes()` promoted to eager_load via `.references()`) builds a JOIN through the auto-generated HABTM join model.

Previously HABTM fell through to `return null` in `addAssociation`, leaving only the preloader path (`.includes` / `.preload`) functional. The join-model PK is composite (`[ownerFk, targetFk]`), but `_addThroughAssociation` only needs the through model's table — not its PK — for HABTM since the source is always `belongsTo`.

### Changes

- `join-dependency.ts`: extend `addAssociation` to accept `hasAndBelongsToMany` and route to `_addThroughAssociation` (HABTM always has `options.through` set by the builder)
- Coerce `assocType` to `"hasMany"` on HABTM JoinNodes so downstream consumers (limit suppression, singular-vs-collection assignment) treat them uniformly

### Tests

- `eager association loading with explicit join habtm` — `.eagerLoad("habtmAssoc")` builds JOIN and reconstructs preloaded collection
- `eager association loading with habtm via preload` — `.preload(...)` path remains green

Existing `.includes()`-based HABTM tests (which run through the preloader) continue to pass.

## Test plan

- [x] `pnpm vitest run packages/activerecord/src/associations/eager.test.ts`
- [x] `pnpm vitest run packages/activerecord/src/associations/habtm.test.ts has-and-belongs-to-many-associations.test.ts inner-join-association.test.ts`

## Followups (out of scope)

- Schema Slot H-b: dedicated `_addHabtmAssociation` only justified if `_addThroughAssociation` reuse becomes brittle — current routing is sufficient.
- Aliased table-name references in `.where()` clauses (HABTM target always aliased to `tN`); broader pre-existing issue not specific to HABTM.

### DIFF
--- a/packages/activerecord/src/associations/eager.test.ts
+++ b/packages/activerecord/src/associations/eager.test.ts
@@ -1647,6 +1647,84 @@ describe("EagerAssociationTest", () => {
     expect(profile).not.toBeNull();
     expect(profile.bio).toBe("HoBio");
   });
+  it("eager association loading with explicit join habtm", async () => {
+    class EjHabtmPost extends Base {
+      static {
+        this.attribute("title", "string");
+        this.adapter = adapter;
+      }
+    }
+    class EjHabtmCategory extends Base {
+      static {
+        this.attribute("name", "string");
+        this.adapter = adapter;
+      }
+    }
+    Associations.hasAndBelongsToMany.call(EjHabtmPost, "ejHabtmCategories", {
+      className: "EjHabtmCategory",
+      joinTable: "ej_habtm_categories_ej_habtm_posts",
+    });
+    registerModel(EjHabtmPost);
+    registerModel(EjHabtmCategory);
+
+    const p1 = await EjHabtmPost.create({ title: "P1" });
+    const p2 = await EjHabtmPost.create({ title: "P2" });
+    const tech = await EjHabtmCategory.create({ name: "Technology" });
+    const gen = await EjHabtmCategory.create({ name: "General" });
+
+    const { CollectionProxy } = await import("./collection-proxy.js");
+    const assoc = (EjHabtmPost as any)._associations.find(
+      (a: any) => a.name === "ejHabtmCategories",
+    )!;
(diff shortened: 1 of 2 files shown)


deanmarano:
## Copilot review #2 response

**Main comment (JOIN path proof):** Addressed — the explicit-eagerLoad test now asserts `toSql()` contains both `LEFT OUTER JOIN ... ej_habtm_categories_ej_habtm_posts` (through/join table) and `LEFT OUTER JOIN ... ej_habtm_categories` (target table). Without the production change the SQL would not contain either, since `addAssociation` previously returned `null` and `_executeEagerLoad` would fall through to a plain SELECT + preload (no JOINs in the SQL at all).

**Suppressed comment (primaryKey option on HABTM):** Repeat of review #1. Still ignored: Rails' `has_and_belongs_to_many` macro (`vendor/rails/.../associations.rb:1900`) only forwards `[:before_add, :after_add, :before_remove, :after_remove, :autosave, :validate, :join_table, :class_name, :extend, :strict_loading]` to the underlying has_many :through. `:primary_key` is intentionally not forwarded by Rails.

**Suppressed comment (schema-qualified HABTM tables):** Pre-existing across all non-HABTM through associations (`_addThroughAssociation` was already quoting full table names in PR-prior code). Not introduced by this PR; out of scope.

**Suppressed comment (limit test could pass via fallback):** The added SQL assertion on the explicit-eagerLoad test now rules out the fallback path for the JOIN wiring; the limit test exercises the same wiring with a LIMIT applied. Adding a SQL assertion to the limit test specifically would only re-prove what the SQL assertion above already proves.

**Suppressed comment (`includes(...).references(...)` promotion):** Repeat of review #1. Out of scope — HABTM target tables alias to `tN` in `_addThroughAssociation`, so user-written `.where("target_table.col")` predicates can't resolve. Pre-existing aliasing concern affecting all through associations.

PR #2158, opened 2026-05-21
docs(tm-unification): capture 9b-1/9b-2a closures + 9b-2c followup

deanmarano:
## Summary

- 9b-1 (#2139) and 9b-2a (#2144) merged since the last refresh; mark them closed.
- 9b-2 is now structured as 9b-2a (Table.star fix) + 9b-2b (gate flip + sweep, in flight as #2155 with split-out fixes #2156 and #2157).
- Adds a **9b-2c follow-up** entry: #2157 swapped `UnqualifiedColumn(table.get(col))` → `SqlLiteral(adapter.quoteColumnName(col))` to fix subquery-alias ORDER BY under MySQL. Functionally equivalent SQL, but bypasses Rails' \`arel_column(field) { |name| quote_table_name(name) }\` yield-block path. Rails never constructs \`UnqualifiedColumn\` for ORDER BY — that node is reserved for UPDATE-set semantics. Restore the Rails-shape post-9b-2b.

## Test plan

- [x] Doc-only change

### DIFF
--- a/docs/tm-unification-plan.md
+++ b/docs/tm-unification-plan.md
@@ -102,10 +102,10 @@ logic:
 
 ### Phase 9a — activate `Visitors::SQLite`, delete `fixSqliteCompat` — closed (#2127, #2132)
 
-- `SchemaAdapter.arelVisitor` (test-adapter.ts:709-719) now delegates
-  to the inner adapter **for SQLite only**. The gate at line 718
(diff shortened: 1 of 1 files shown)


deanmarano:
Doc refresh wasn't requested — closing without merge. Plan-doc updates will be folded into specific Phase 9 PRs going forward.

PR #6748, opened 2026-08-19
refactor(activerecord): derive CollectionProxy's delegate-to-scope list from mixin keys

deanmarano:
## Context

`CollectionProxy` delegated to `scope()` using two hand-transcribed name
arrays standing in for `QueryMethods`/`SpawnMethods`'
`public_instance_methods(false)`
(`vendor/rails/activerecord/lib/active_record/associations/collection_proxy.rb:1128-1137`).
`QueryMethodBangs` and `SpawnMethods` are the exact mixin objects
`include()`'d into `Relation` (`relation.ts:3624-3628`), so `Object.keys()`
on them IS that reflection.

## Change

- `SPAWN_METHODS_PUBLIC_INSTANCE_METHODS` and `QUERY_METHODS_PUBLIC_INSTANCE_METHODS`
  are now derived from `Object.keys(SpawnMethods)` / `Object.keys(QueryMethodBangs)`,
  excluding `query_methods.rb`'s `protected`/`private` sections
  (`query_methods.rb:1603`, `:1677`) and `spawn_methods.rb`'s `private def
  relation_with` (`spawn_methods.rb:71`) by name — none of those are in
  `public_instance_methods(false)`.
- `QueryMethodBangs` turned out to already carry trails' *entire* `QueryMethods`
  surface (bang and non-bang), not just the bang half the story anticipated —
  RFC 0107's fan-out had progressed further than the story's context assumed.
  So the whole delegate-relevant surface is now derived; no hand-transcribed
  residual remains.
- New test `collection-proxy-delegate-methods.trails.test.ts` pins the derived
  `delegateMethods` set and asserts no `query_methods.rb` protected/private
  helper is ever delegated.

## Byte-identical scope

The derivation surfaced real drift between the old hand-list and Rails:
`leftJoins`/`without` (public aliases, `query_methods.rb:887,1585`) were never
delegated, and `nullBang`/`rewhereBang`/`selectBang` were dead names matching
no real Rails or trails method. Per review, those fixes are out of this
story's scope — its AC pins the delegated set byte-identical to before. This
PR now explicitly excludes `leftJoins`/`without` from the derived set and
keeps the three dead names in the extra-name concat (both with call-site
comments), so the set is unchanged. The fixes are tracked in
`collection-proxy-delegate-leftjoins-without-fix`.

`pnpm parity:api:calls` / `:args` add zero rows for this file.
`pnpm parity:api:extra` reports no new surface from this change.

Closes-story: delegate-list-from-mixin-keys-bakeoff-sonnet


deanmarano:
Closing: bake-off arm (bakeoff-do-not-merge), experiment concluded — see bakeoff-audit.md.

PR #6794, opened 2026-08-20
refactor(activemodel): converge resolveAliasNameIn onto resolveAttributeName

deanmarano:
`resolveAliasNameIn` was the last extra alias-resolution spelling in activemodel: a two-argument (`present`) variant applying a trails-only camelCase-key bridge on top of Rails' plain `attribute_aliases[name] || name`.

Rails resolves aliases with one identical step in `read_attribute` (read.rb:31-34), `write_attribute` (write.rb:31-34) and `has_attribute?` (attribute_methods.rb:316-319), spelled `resolve_attribute_name` at attribute_methods.rb:396-398. Every call site — activemodel `model.ts` (read/write/hasAttribute), `activerecord/base.ts` class-level `hasAttribute`, `readonly-attributes.ts`, and AR's `attribute-methods.ts` — now goes through `resolveAttributeName`. AR's `recordAliasName` wrapper existed only to share the bridge and is gone with it.

The four trails-only tests in `attribute-methods.trails.test.ts` that asserted the bridge's guess-and-ordering semantics are removed; they guarded behavior Rails does not have. No Rails-mirroring test depended on the bridge.

`parity:api:extra --package activemodel`: index.ts 24 → 23 novel. Call/arg ratchets OK; activemodel suite and the AR attribute/alias/counter-cache/dirty/timestamp files pass.

Closes-story: converge-resolve-alias-name-in-onto-resolve-attribute-name


deanmarano:
Superseded: #6781 landed this exact convergence (resolveAliasNameIn and the camelCase alias bridge removed; all call sites on resolveAttributeName). Rebasing this branch onto main leaves an empty diff.

